PR #20, opened 2021-08-10
WIP: testing various network topologies

jtieri:
Working on various network configurations/topologies


cmd/horcrux/cmd/cosigner.go
```diff
@@ -45,7 +55,19 @@ func StartCosignerCmd() *cobra.Command {
 				pv       types.PrivValidator
 				chainID  = config.ChainID
 				logger   = tmlog.NewTMLogger(tmlog.NewSyncWriter(os.Stdout)).With("module", "validator")
-				config   = signer.Config{
+				cfg      signer.Config
+			)
+
+			if single {
```
jackzampolin, line 61: Thank you for catching this bone headed move. One easy refactor will be to remove the `signer.Config` object from the `signer/` lib and use the one here instead

cmd/horcrux/cmd/cosigner.go
```diff
@@ -55,102 +77,117 @@ func StartCosignerCmd() *cobra.Command {
...
-			if err != nil {
-				panic(err)
-			}
+			var val types.PrivValidator
+			if single {
+				stateFile := path.Join(cfg.PrivValStateDir, fmt.Sprintf("%s_priv_validator_state.json", chainID))
 
-			cosigners := []signer.Cosigner{}
-			remoteCosigners := []signer.RemoteCosigner{}
+				// TODO either stop creating state file at config init
```
jackzampolin, line 92: Either stop creating state file at config init, or document the hell out of this. Signer cluster will sign whatever block the node throws at it so state file is only protection on this side from doublesign. Not creating on init and having a separate command tree would raise user awareness of the potential doublesign dangers here. Commands for management of this state are nice to haves as well.

cmd/horcrux/cmd/cosigner.go
```diff
@@ -55,102 +77,117 @@ func StartCosignerCmd() *cobra.Command {
...
-			total := len(config.Cosigners) + 1
-			localCosignerConfig := signer.LocalCosignerConfig{
-				CosignerKey: key,
-				SignState:   &shareSignState,
-				RsaKey:      key.RSAKey,
-				Peers:       peers,
-				Total:       uint8(total),
-				Threshold:   uint8(config.CosignerThreshold),
-			}
+				// add ourselves as a peer so localcosigner can handle GetEphSecPart requests
```
jackzampolin, line 126: This is an interesting comment.

cmd/horcrux/cmd/cosigner.go
jtieri: State file will still need to be generated with config init, for now. Adding commands to fetch and manage state is a good idea, something to make note of and get done in another PR.

PR #22, opened 2021-08-16
Change `--config` flag to `--home`

jtieri:
Instead of passing the full path to the config file, e.g. `~/.horcrux/config.yaml`, the directory is passed in and the config file is inferred from there

### DIFF
--- a/cmd/horcrux/cmd/root.go
+++ b/cmd/horcrux/cmd/root.go
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	cfgFile string
+	homeDir string
 	config  *Config
 )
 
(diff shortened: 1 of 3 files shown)


cmd/horcrux/cmd/root.go
```diff
@@ -30,13 +30,14 @@ func Execute() {
 
 func init() {
 	cobra.OnInitialize(initConfig)
-	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.horcrux/config.yaml)")
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "home", "", "directory for config and data (default is $HOME/.horcrux/)")
```
jackzampolin, line 33: One note here: we are using the `cfgFile` var to store the home. This could be confusing. Maybe change the `cfgFile` var to `homeDir` or something?

cmd/horcrux/cmd/root.go
```diff
@@ -30,13 +30,14 @@ func Execute() {
...
 }
 
 // initConfig reads in config file and ENV variables if set.
 func initConfig() {
 	cfg := Config{}
 	if cfgFile != "" {
+		cfgFile = path.Join(cfgFile, "config.yaml")
```
jackzampolin, line 40: If you make the above change, then this will need to change too.

jackzampolin:
Just that one comment and once tests are passing again please feel free to merge!

cmd/horcrux/cmd/root.go
jtieri: I've tweaked things now where `homeDir` will point to the actual directory but this part will remain `cfgFile` since it actually points to the `config.yaml` file

cmd/horcrux/cmd/root.go
jtieri: Yeah that makes sense, I forgot to leave a note and I didn't change the name due to having to tweak some statements in `config.go`

I went ahead and refactored it and changed some parts in `test/signer_test.go` and `config.go`so that `--home` functions as intended

Issue #25, opened 2021-09-02
Cosigner RPC Server Configurable Timeout


ryanh-labs:
I think this project is fantastic and am a huge fan of everything happening here in the way of improvements and useability 😃 .

Another improvement I think would be awesome is to make the timeout for the cosigner rpc server configurable (currently hardcoded to 1 second). Depending on the latency between the cosigners a value of 1200-1500ms drastically reduces the number of missed signatures for cosigners located in different regions/DCs.

https://github.com/strangelove-ventures/horcrux/blob/547bf1b00c6169b410f1e731427067662c566c32/signer/cosigner_rpc_server.go#L127

PR #26, opened 2021-09-02
Make Cosigner RPC Server Timeout Value Configurable

jtieri:
The proposed changes would make the Cosigner RPC server have a configurable timeout value (as requested in #25 ) that can be set with `horcrux cosigner start --timeout str`  and when no flag is provided the default value is set to 1 second.

The accepted values for `--timeout str` are those that are accepted by Go's `time.ParseDuration()` function

From [https://pkg.go.dev/time#Duration](url)
```
ParseDuration parses a duration string. A duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```

Edit: Added this to `cmd/config` instead. The flag will now instead take in an int representing the timeout value in milliseconds with a default value of 1500ms

### DIFF
--- a/cmd/horcrux/cmd/config.go
+++ b/cmd/horcrux/cmd/config.go
@@ -126,7 +126,7 @@ func initCmd() *cobra.Command {
 	cmd.Flags().StringP("peers", "p", "", "cosigner peer addresses in format tcp://{addr}:{port}|{share-id} (i.e. \"tcp://node-1:2222|2,tcp://node-2:2222|3\")")
 	cmd.Flags().IntP("threshold", "t", 0, "indicate number of signatures required for threshold signature")
 	cmd.Flags().StringP("listen", "l", "tcp://0.0.0.0:2222", "listen address of the signer")
-	cmd.Flags().Int("timeout", 1000, "configure cosigner rpc server timeout value, "+
+	cmd.Flags().Int("timeout", 1500, "configure cosigner rpc server timeout value, "+
 		"units are in milliseconds e.g. 1000 = 1 second ")
 	return cmd
 }
@@ -161,9 +161,9 @@ func validateCosignerConfig(cfg *Config) error {
 	if cfg.CosignerConfig.Threshold < 2 {
 		return fmt.Errorf("threshold must be 2 or greater")
 	}
-	if cfg.CosignerConfig.Timeout < 1 || cfg.CosignerConfig.Timeout > 1500 {
-		return fmt.Errorf("(%d) is an invalid timeout value. "+
-			"cosigner RPC server timeout must be greater than 0ms and less than 1500ms", cfg.CosignerConfig.Timeout)
+	if cfg.CosignerConfig.Timeout < 1 {
+		return fmt.Errorf("(%d) is an invalid timeout value. cosigner RPC server timeout must be greater than 0ms ",
+			cfg.CosignerConfig.Timeout)
 	}
 	if _, err := url.Parse(cfg.CosignerConfig.P2PListen); err != nil {
 		return fmt.Errorf("failed to parse p2p listen address")
--- a/cmd/horcrux/cmd/config.go
+++ b/cmd/horcrux/cmd/config.go
@@ -67,6 +67,8 @@ func initCmd() *cobra.Command {
 				threshold, _ := cmd.Flags().GetInt("threshold")
 				peers, err := peersFromFlag(p)
 				listen, _ := cmd.Flags().GetString("listen")
+				timeout, _ := cmd.Flags().GetInt("timeout")
+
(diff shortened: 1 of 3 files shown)


cmd/horcrux/cmd/cosigner.go
```diff
@@ -30,12 +30,14 @@ var cosignerCmd = &cobra.Command{
 
 func StartCosignerCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "start [single-signer]",
+		Use:   "start [single-signer] [rpc-timeout]",
```
jackzampolin, line 33: because both of these are flags, I don't think we need to put them as `args` here

cmd/horcrux/cmd/cosigner.go
```diff
@@ -30,12 +30,14 @@ var cosignerCmd = &cobra.Command{
...
 		Short: "start cosigner process",
-		Long:  "[single-signer] specifies that horcrux should be ran as a single signer",
-		Args:  cobra.NoArgs,
+		Long: "[single-signer] specifies that horcrux should be ran as a single signer \n" +
+			"[rpc-timout] specifies the timeout for the cosigner rpc server ",
+		Args: cobra.RangeArgs(0, 2),
```
jackzampolin, line 37:
```suggestion
		Args: cobra.NoArgs
```

cmd/horcrux/cmd/cosigner.go
```diff
@@ -215,5 +222,7 @@ func StartCosignerCmd() *cobra.Command {
 	}
 
 	cmd.Flags().BoolP("single", "s", false, "set to start horcrux as a single signer")
+	cmd.Flags().StringP("timeout", "t", "1s", "configure cosigner rpc server timeout value, "+
```
jackzampolin, line 225: Ideally we add this to the config file and set a sane default (1500ms) when running `config init`

signer/cosigner_rpc_server.go
```diff
@@ -124,7 +127,7 @@ func (rpcServer *CosignerRpcServer) rpcSignRequest(ctx *rpc_types.Context, req R
 
 			// RPC requests are blocking
 			// to prevent it from hanging our process indefinitely, we use a timeout context and a goroutine
-			partReqCtx, partReqCtxCancel := context.WithTimeout(context.Background(), time.Second)
+			partReqCtx, partReqCtxCancel := context.WithTimeout(context.Background(), rpcServer.timeout)
```
jackzampolin, line 130: perfect!

jackzampolin:
Got the timeout down to the calls, perfect. But I think we should put this into the config, not into command line args and flags.

cmd/horcrux/cmd/cosigner.go
jtieri: Done! Added this to `cmd/config` instead and took your above recommended changes into account while doing so.
The flag will take in an int  representing the timeout value in milliseconds with a default value of 1500ms

cmd/horcrux/cmd/config.go
```diff
@@ -191,9 +200,10 @@ func (c *Config) MustMarshalYaml() []byte {
 }
 
 type CosignerConfig struct {
-	Threshold int            `json:"threshold" yaml:"threshold"`
-	P2PListen string         `json:"p2p-listen" yaml:"p2p-listen"`
-	Peers     []CosignerPeer `json:"peers" yaml:"peers"`
+	Threshold int            `json:"threshold"   yaml:"threshold"`
+	P2PListen string         `json:"p2p-listen"  yaml:"p2p-listen"`
+	Peers     []CosignerPeer `json:"peers"       yaml:"peers"`
+	Timeout   int            `json:"rpc-timeout" yaml:"rpc-timeout"`
```
jackzampolin, line 206: Boom

jackzampolin: lets set this as a string and use `time.ParseDuration` to parse it

cmd/horcrux/cmd/cosigner.go
```diff
@@ -165,11 +163,17 @@ func StartCosignerCmd() *cobra.Command {
 					Peers:     cosigners,
 				})
 
+				duration := fmt.Sprintf("%d%s", config.CosignerConfig.Timeout, "ms")
```
jackzampolin, line 166: so instead of setting it as a `ms` value by default, let user put in string of their choosing and in the config validation logic parse it as a duration

jtieri: Consider it done! :100:

Issue #18, opened 2021-08-08
Commands to read and modify config and state files


jackzampolin:
Add commands to the CLI like so:

```
horcrux
  config 
    init
    nodes
      add
      remove
    peers
      add
      remove
  state
    pv
      show
      reset (--height)
    share
      show
      reset (--height)
```

freshfab:
Are you guys working on this? If not, I'd love to help with this one. 😄

PR #31, opened 2021-11-18
Add commands to read and modify config and state files

freshfab:
This PR addresses #18.

- [X] `config nodes add/remove`
- [X] `config peers add/remove`
- [X] `config chain-id set`
- [X] `state pv show/reset (--height)`
- [X] `state share show/reset (--height)`
- [X] Tests

### Notes

I figured resetting either `pv` or `share` should only be possible if horcrux is not running since they are constantly being written to, so I put a unix-style cli check in there (for lack of a better go-style solution), which comes with a couple of caveats:

1. The check only works on UNIX-based systems.
2. The binary cannot be renamed to anything other than `horcrux`.

### DIFF
--- a/cmd/horcrux/cmd/config.go
+++ b/cmd/horcrux/cmd/config.go
@@ -19,8 +19,8 @@ import (
 
 func init() {
 	// TODO: config nodes add/remove
-	nodesCmd.AddCommand(addCmd())
-	nodesCmd.AddCommand(removeCmd())
+	nodesCmd.AddCommand(addNodesCmd())
+	nodesCmd.AddCommand(removeNodesCmd())
 	configCmd.AddCommand(nodesCmd)
 	// TODO: config peers add/remove
 	// TODO: config chain-id set
@@ -188,7 +188,7 @@ var nodesCmd = &cobra.Command{
 	Short: "Commands to configure the chain nodes",
 }
 
-func addCmd() *cobra.Command {
+func addNodesCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:     "add [chain-nodes]",
 		Aliases: []string{"a"},
@@ -229,7 +229,7 @@ func addCmd() *cobra.Command {
 	}
 }
 
-func removeCmd() *cobra.Command {
+func removeNodesCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:     "remove [chain-nodes]",
 		Aliases: []string{"r"},
--- a/cmd/horcrux/cmd/config.go
+++ b/cmd/horcrux/cmd/config.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -484,37 +483,6 @@ type CosignerPeer struct {
 	P2PAddr string `json:"p2p-addr" yaml:"p2p-addr"`
 }
 
-func readKeyShare() (*signer.CosignerKey, error) {
-	// Read ID from local key share
-	var home string // In root.go we end up with our
-	if homeDir != "" {
-		home = homeDir
-	} else {
-		home, _ = homedir.Dir()
-		home = path.Join(home, ".horcrux")
-	}
-
-	if _, err := os.Stat(homeDir); !os.IsNotExist(err) {
-		return nil, fmt.Errorf("%s is not empty, check for existing configuration and clear path before trying again", homeDir)
-	}
-
-	file, err := os.Open(path.Join(home, "share.json"))
-	if err != nil {
-		return nil, err
-	}
-	defer file.Close()
-
-	bz, err := ioutil.ReadAll(file)
-	if err != nil {
-		return nil, err
-	}
-	var key signer.CosignerKey
-	if err := json.Unmarshal(bz, &key); err != nil {
-		return nil, err
-	}
-	return &key, nil
-}
-
 func validateCosignerPeers(peers []CosignerPeer) error {
 	// Check IDs to make sure none are duplicated
 	if dupl := duplicatePeers(peers); len(dupl) != 0 {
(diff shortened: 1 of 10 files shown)


cmd/horcrux/cmd/config.go
```diff
@@ -218,13 +484,69 @@ type CosignerPeer struct {
...
 func validateCosignerPeers(peers []CosignerPeer) error {
-	// TODO: check IDs to ensure none are duplicated and
-	// that the key share which is configured for the local node
+	// Check IDs to make sure none are duplicated
+	if dupl := duplicatePeers(peers); len(dupl) != 0 {
+		return fmt.Errorf("found duplicate share IDs in args: %v", dupl)
+	}
+
+	// Check that the key share which is configured for the local node
 	// is the last remaining share
```
freshfab, line 525: My initial approach was to just go through the peers in the config and check for the missing ID, which assumes that IDs are increments of 1. Since I don't know if that's safe to assume, I just read the share ID from the `share.json`.

However, that means that the `share.json` needs to be present before I can use `horcrux config init`, which requires creating the config directory manually, so that's obviously pointless.

So my question is, is it safe to assume that the share IDs are always increments of 1, starting from ID 1?

jackzampolin:
This is huge @freshfab thank you! Will take a look

cmd/horcrux/cmd/config.go
jackzampolin: Yes it is safe to assume that share id's are 1 indexed and incrementing up from 1

cmd/horcrux/cmd/config.go
```diff
@@ -177,6 +186,263 @@ func validateCosignerConfig(cfg *Config) error {
...
+			}
+			if _, err := os.Stat(homeDir); !os.IsNotExist(err) {
+				return fmt.Errorf("%s is not empty, check for existing configuration and clear path before trying again", homeDir)
+			}
+
+			stateDir := path.Join(home, "state")
+			pvOldPath := path.Join(stateDir, config.ChainID+"_priv_validator_state.json")
+			pvNewPath := path.Join(stateDir, args[0]+"_priv_validator_state.json")
+			shareOldPath := path.Join(stateDir, config.ChainID+"_share_sign_state.json")
+			shareNewPath := path.Join(stateDir, args[0]+"_share_sign_state.json")
```
jackzampolin, line 428: Nice

cmd/horcrux/cmd/config.go
```diff
@@ -218,13 +484,69 @@ type CosignerPeer struct {
...
+	// Check that the key share which is configured for the local node
 	// is the last remaining share
+	key, err := readKeyShare()
+	if err != nil {
+		return err
+	}
+	for _, peer := range config.CosignerPeers() {
+		if peer.ID == key.ID {
+			return fmt.Errorf("peer with share ID %v cannot be added", key.ID)
+		}
```
jackzampolin, line 533: Great safety check

cmd/horcrux/cmd/state.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+func init() {
+	pvCmd.AddCommand(showPvCmd())
+	pvCmd.AddCommand(resetPvCmd())
+	stateCmd.AddCommand(pvCmd)
+
+	shareCmd.AddCommand(showShareCmd())
+	shareCmd.AddCommand(resetShareCmd())
+	stateCmd.AddCommand(shareCmd)
```
jackzampolin, line 23: I think you can merge these two. I don't see a usecase for individually setting them so lets make them just 1 command to set the state for both.

jackzampolin:
A couple of comments @freshfab but this is on the right track!

cmd/horcrux/cmd/state.go
jackzampolin: Also lets make the `--height` a required arg as opposed to a flag.

freshfab: Gotcha, I've changed it to the following:
```bash
horcrux
    state
        show
        set [height]
```

cmd/horcrux/cmd/config.go
freshfab: I'm open to suggestions 😄

cmd/horcrux/cmd/config.go
freshfab: As far as I understand, I need the max number of peers to make this check, like if I have 3 peers, and IDs 1 and 2 are already present, the local node is obviously ID 3. I could use some pointers here, if you don't mind.

cmd/horcrux/cmd/state.go
jackzampolin: This is awesome!

cmd/horcrux/cmd/config.go
jackzampolin: I was thinking option 1: i.e. assume that the peers in the config are correct and that the node we are configuring is the missing ID

cmd/horcrux/cmd/config.go
jackzampolin: I like this. Its good as is

cmd/horcrux/cmd/config.go
freshfab: Ok, here's what I did:
```bash
horcrux config peers set-shares [num-shares]
```
This adds a new field `shares` to the cosigner config, which sets the boundaries for adding new nodes/peers to the config.

freshfab:
@jackzampolin  I've finished up the unit tests for `config` and `state`. If there are any more test cases you'd like to cover, please let me know.

jackzampolin:
Huge! This looks amazing

jackzampolin:
Looks like there are some failing test cases because of the command format changes. Would you mind taking a look at those?

jackzampolin:
The only test failure is as follows:

```
=== RUN   TestStateSetCmd/valid_height
    state_test.go:69: 
        	Error Trace:	state_test.go:69
        	Error:      	Received unexpected error:
        	            	open /tmp/TestStateSetCmd/.horcrux/state/horcrux-1_priv_validator_state.json: no such file or directory
        	Test:       	TestStateSetCmd/valid_height
```

freshfab:
Took me a while to figure out that `go-homedir` caching interfered with `horcrux`'s caching, so I disabled that for the tests. All tests should be working now.

jackzampolin:
Thank you @freshfab !

Issue #29, opened 2021-11-10
Add test case for downed signer


jackzampolin:
We need to simulate taking down individual nodes in the signer cluster and ensure that the HA is working as intended. Maybe test takes down each signer for a period of time and ensures continued operation

mturkia:
When horcrux traffic is dropped between the cluster nodes, this leads to missing signing. Can be replicated e.g. with_iptables -I INPUT -p tcp --dport <p2p-listen port> -j DROP_
I initially noticed this when rebooting one horcrux node and missing signing during that time.

PR #32, opened 2021-11-23
Refactor StartCosginerContainers() + add facilities for testing downed signers

jtieri:
Previously `StartCosignerContainers()` worked off the assumption that there were an even number of sentry nodes per signer node (e.g. 3 signer nodes with 2 sentries per node required 6 sentry nodes)

This refactor removes that assumption such that the appropriate number of sentries will be assigned per signer node in the cluster regardless of how many sentry nodes are present (i.e. any arbitrary number of sentry nodes will work)

Also added some functions needed to pause/unpause a signer node's docker container along with the test to test downed signers in a cluster as per #29, however, the test currently fails with indeterminate results. After running the test several times I received results anywhere between 1-4 blocks missed with the first one or two downed signers occasionally not missing any blocks. Gonna skip it in CI right now and circle back to finding the cause for failure/a fix in another PR


jackzampolin:
When running this I've been getting errors due to failed `gentx` containers

jackzampolin:
ok i figured out the gentx issue i've been getting. due to race condition and unneeded concurrency around file access for genesis creation. going to push that change here and abandon my other branch.

jackzampolin (approved):

jackzampolin:
Merging this PR, after exploration in #34 this will likely be the last change to the current version of Horcrux. We are going to replace the current arch with a raft based cluster for leader election. Great work by @jtieri and @agouin finding the core issues with the current p2p network between the signers.

PR #38, opened 2022-01-12
Raft integration

### LINKED ISSUE #29: Add test case for downed signer

### LINKED ISSUE #30: Proposal block can never be committed
With `horcrux` setup correctly, the validator's proposal block will always get the following info and the block will never be committed, resulting selecting a new proposer and slow down the speed of chain.

```
INF commit is for a block we do not know about; set ProposalBlock=nil commit=C52CB815E75718494DA41B0D49006B7E90597BE39B8A3FF17215596E148F0FC2 commit_round=1 height=2900785 module=consensus proposal=
```
The following is the log from the signer. We can see that another precommit is voted at `round=1`.

```
Nov 15 05:06:16 desmos-m-signer-3-ln-sg horcrux[153734]: I[2021-11-15|05:06:16.569] Signed proposal                              module=validator node={validator_address} height=2900785 round=0 type=SIGNED_MSG_TYPE_PROPOSAL
Nov 15 05:06:17 desmos-m-signer-3-ln-sg horcrux[153734]: I[2021-11-15|05:06:17.850] Signed vote                                  module=validator node={validator_address} height=2900785 round=0 type=SIGNED_MSG_TYPE_PREVOTE
Nov 15 05:06:19 desmos-m-signer-3-ln-sg horcrux[153734]: I[2021-11-15|05:06:19.131] Signed vote                                  module=validator node={validator_address} height=2900785 round=0 type=SIGNED_MSG_TYPE_PRECOMMIT
Nov 15 05:06:20 desmos-m-signer-3-ln-sg horcrux[153734]: I[2021-11-15|05:06:20.433] Signed vote                                  module=validator node={validator_address} height=2900785 round=1 type=SIGNED_MSG_TYPE_PRECOMMIT
```

### LINKED ISSUE #36: Horcrux v2 - Raft edition 🛶
We are going to implement a raft cluster for leader election. Only the leader will be able to submit signatures to sentries and all nodes will still maintain a high watermark HRS file outside consensus for protection against node take over (i.e. in the case where an attacker gains access to a single node in the cluster the attacker would only be able to cause a service disruption not a double sign event).

We plan to use https://github.com/strangelove-ventures/raft-sample as a basis for the implementation i.e. using [hashicorp/raft](https://github.com/hashicorp/raft) and a database backend ([badgerdb](https://github.com/dgraph-io/badger)?). Data to be stored in consensus:
- High watermark (i.e. what is the latest HRS that the cluster has sent to sentries)
- Sentry nodes
- Cluster members?

agouin:
Introduces the raft protocol to the cluster for the purposes of leader election and high watermark consensus among signer nodes

All sign requests from the sentries made to any signer node will now be proxied through the raft leader. 

The signer node that is the current elected raft leader will act upon the sign requests by managing the threshold validation process:

- Check the requested block against the high watermark file (kept in consensus between the signer nodes) to avoid double signing.
- Request ephemeral nonces for the block signature from each signer node peer.
- Each signer will act upon the request by generating the ephemeral nonce shares for all other signers (encrypted with the destination signer's RSA public key). These shares will be the response to the leader.
- The leader will wait until it has received _`t - 1`_ responses. The signer nodes which responded in time, _`blockSigners`_ are the signers that will be included with the leader for signing the block.
- The leader will then make a request to each of the _`blockSigners`_ to set the ephemeral nonces for the other signers that are participating in the block signing (_`blockSigners`_ and leader), and produce the signature part from the block data.
- The participant in _`blockSigners`_ will handle this request by decrypting the ephemeral shares with its RSA private key, verify the signatures of the ephemeral share to verify the identity of the source signers, and then save it in memory. After all of the nonces are saved (consensus with the leader and _`blockSigners`_), it will produce its signature piece for the block data and respond to the leader with it.
- Once the leader receives the signature parts from all of the _`blockSigners`_, it will make a combined signature including its own signature part and those from the _`blockSigners`_
- The leader will verify the combined signature is valid, then update its own high watermark file and also emit the block metadata (height, round, and step), to the rest of the signers through raft in order to update their high watermark files. This gives the cluster consensus on what the last successfully signed block was.
- The leader will finally respond with the combined signature for the block, either directly to the requesting sentry if the raft leader was the one who handled the sentry request, or the signer that proxied the request to the leader, which would then respond to the requesting sentry.

The cluster can now reliably sign when (`total signers` - `threshold signers`) are down. If a signer node goes down and it is not the leader, there should not be any missed blocks. If the leader goes down, a new leader will be elected and it will pick up signing blocks after that, so there might be a few missed blocks in the case of a downed leader.

Requires a configuration change to `config.yaml` on the signer nodes to indicate the raft listen address and the raft address of the cosigners.
For example:

```yaml
chain-id: horcrux
cosigner:
  threshold: 3
  shares: 5
  p2p-listen: tcp://0.0.0.0:2222
  raft-listen: signer-5-TestDownedSigners3of5:2223
  peers:
  - share-id: 1
    p2p-addr: tcp://signer-1-TestDownedSigners3of5:2222
    raft-addr: signer-1-TestDownedSigners3of5:2223
  - share-id: 2
    p2p-addr: tcp://signer-2-TestDownedSigners3of5:2222
    raft-addr: signer-2-TestDownedSigners3of5:2223
  - share-id: 3
    p2p-addr: tcp://signer-3-TestDownedSigners3of5:2222
    raft-addr: signer-3-TestDownedSigners3of5:2223
  - share-id: 4
    p2p-addr: tcp://signer-4-TestDownedSigners3of5:2222
    raft-addr: signer-4-TestDownedSigners3of5:2223
  rpc-timeout: 1500ms
chain-nodes:
- priv-val-addr: node-0-TestDownedSigners3of5:1234
```

This configuration can be generated with the `horcrux config init` command by providing a new flag `--raft`/`-r` with the raft listen address, and providing the peer raft addresses to the existing `--peers`/`-p` flag. 

The raft listen address needs to be the address where the node can be discovered by other signers (e.g. `signer-5-TestDownedSigners3of5:2223` and not `0.0.0.0:2223`

The peers flag has been modified, and now also requires the raft port to be provided.
The format is csv, where each peer is `tcp://<peer hostname>:<p2p port>|<raft port>|<signer ID>` For example: `tcp://node-1:2222|2223|2,tcp://node-2:2222|2223|3`

Demo:
[Demo](https://www.youtube.com/watch?v=O-yy1CYBDsI)

Benchmarks:
Screenshot from 2022-01-31 13-50-36

Closes #36, Closes #30, Closes #29


jackzampolin (approved):
Looking great @agouin, giving this a +1 so that we can merge into master whenever. Once you are feeling confident in the testing we can cut this as a v2.0.0-beta1 or something like that.

PR #44, opened 2022-02-01
Fix race condition when initial leader is brought back up

agouin:
Simplify cluster bring-up
No initial leader needs to be defined. The raft BootstrapCluster method can be called by non leaders, and errors can be ignored. By passing all peers to the initial raft configuration, no join calls are necessary. 
The signer that first starts will then be the initial leader, and it won't have issues when rejoining the cluster

Updates downed signer tests to wait for 3 consecutive successfully signed blocks after taking down a node before starting to check missed blocks from then on to make test assertions more concrete and functional on machines with different resources.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,20 +17,18 @@ jobs:
         with:
           go-version: 1.17
 
-      
       # checkout horcrux
       - name: checkout horcrux
         uses: actions/checkout@v2
 
-      # build cache
-      - uses: actions/cache@v1
-        with:
-          path: ~/go/pkg/mod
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -17,23 +17,21 @@ jobs:
...
-            ${{ runner.os }}-go-
+      # cleanup docker environment on self-hosted test runner
+      - name: prepare fresh docker environment
+        run: |
+          docker stop $(docker ps -a -q) || true
+          docker rm -f $(docker ps -a -q) || true
+          docker network prune -f || true
 
-      # todo: pull image from repo
+      # todo: use heighliner image
```
jackzampolin, line 31: 👀

jackzampolin (approved):

PR #48, opened 2022-02-03
Use sim image from heighliner for simd

agouin:
Add dependency to [heighliner](https://github.com/strangelove-ventures/heighliner) for simd docker image, which has sim building with the latest versions of the cosmos SDK and stored in GHCR.

Bump Cosmos SDK to 0.44.5


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.17
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/avast/retry-go v3.0.0+incompatible
-	github.com/cosmos/cosmos-sdk v0.44.0
+	github.com/cosmos/cosmos-sdk v0.44.5
```
jackzampolin, line 8: This should be good, but lets be very mindful of both SDK and tendermint version bumps.

jackzampolin (approved):

PR #49, opened 2022-02-04
Add ability to override private key file path in config and cli

### LINKED ISSUE #50: Add config property for custom priv key share path
It can be desirable to have the priv key share at a different path than `share.json`, e.g. a different mounted filesystem.

agouin:
Adds a `config.yaml` string property `key_file`.

Adds a `--keyfile` or `-k` flag to `horcrux config init` to be able to set a custom file path for the key/key-share file when generating the `config.yaml`. Documentation for this flag added to migration doc.

Removes confusing toml properties on the internal Config struct.

Fixes typo in systemd example.

Resolves #50


cmd/horcrux/cmd/cosigner.go
```diff
@@ -44,9 +44,16 @@ func StartCosignerCmd() *cobra.Command {
 				cfg      signer.Config
 			)
 
+			var privValKeyFile string
+			if config.PrivValKeyFile == "" {
```
jackzampolin, line 48: kk this is fine

cmd/horcrux/cmd/config.go
```diff
@@ -511,6 +518,7 @@ func setChainIDCmd() *cobra.Command {
 
 type Config struct {
 	HomeDir        string          `json:"home-dir" yaml:"home-dir"`
+	PrivValKeyFile string          `json:"key_file" yaml:"key_file"`
```
jackzampolin, line 521: maybe add `omitempty` here?

signer/Config.go
```diff
@@ -3,44 +3,28 @@ package signer
...
-	PrivValStateDir   string           `toml:"state_dir"`
-	ChainID           string           `toml:"chain_id"`
-	CosignerThreshold int              `toml:"cosigner_threshold"`
-	ListenAddress     string           `toml:"cosigner_listen_address"`
-	RaftListenAddress string           `toml:"cosigner_raft_listen_address"`
-	Nodes             []NodeConfig     `toml:"node"`
-	Cosigners         []CosignerConfig `toml:"cosigner"`
-}
-
-func LoadConfigFromFile(file string) (Config, error) {
```
jackzampolin, line 32: is this no longer used?

agouin: no usages in the project. wasn't a recent change from what I can tell, just dead code that's been in here a while.

cmd/horcrux/cmd/config.go
agouin: done

jackzampolin (approved):

PR #52, opened 2022-02-06
[DO NOT MERGE] Use our raft fork to get leader ID

agouin:
Better than indirect cosigner fix #51, but don't want to maintain a raft fork. Testing this out and will put up a PR into hashicorp/raft


agouin:
No longer necessary with #61

PR #61, opened 2022-03-07
Use GRPC for Raft and Leader calls

### LINKED ISSUE #53: Ability to change the active leader.
Requested Feature: The ability to change the active leader.  

It would be ideal to be issue a command to either rotate to a new leader, or temporarily set the leader. This would be used during maintenance operations to maintain uptime.

### LINKED ISSUE #60: Snapshot creating/reaping flood on node return to set
I don't have proper logs of this, but can attempt to do a better job if it happens again.  IHere's a screenshot:

a60d1085-13e7-4c0e-ac12-db8430e1313a

After I purposefully kill a node (non-leader), and bring that node back in, occasionally it will be a flood of these creating/reaping snapshots across all nodes. I don't believe it's signing during this flood, at least from quick perusal of the logs.  If I remove a node (any of the 3 in this case), it stops.  When I bring a node back, this same snapshot flood continues until a node is killed.

The only way I've been able to make this stop is to force quit all nodes, and bring them up simultaneously. 

If it's helpful: debian 11, 4 cpu, 8gb ram each.  These are all in the same rack of hardware, so 0 latency between nodes.  If I can help troubleshoot this please let me know.

agouin:
- Consolidates P2P port and Raft port to single port for all cluster traffic. Backwards compatible with Horcrux v1 and v2.0.0-beta3 config.
Removes `--raft` flag from `horcrux config init`. Now, the `--listen`/`-l` flag is required, and must be a DNS/IP that is reachable by the other peers.
- Uses GRPC for requests to the leader. Removes prior RPC server entirely.
- Solves issue of having to cross-lookup leader RPC address from mapping of peer raft addresses to RPC addresses. Now the leader raft address can be called directly with the GRPC calls.
- Uses boltdb for raft log and stable stores instead of in-memory store.
- Adds `horcrux elect` command, which can be executed with or without an argument. If an argument is provided, it is the ID of the node that should be elected leader. Otherwise, the next candidate will be elected.
- Allows signing votes for same HRS with distinct timestamp. HRST keys (height, round, step, and timestamp) are now used for ephemeral share caching instead of HRS alone. Proposals have a special case though, the first proposal should always be returned.
- No longer signs votes with nil BlockID

Resolves #53 
Resolves #60


jackzampolin:
This is working well on our testnet validators. Going to give this the 👍

jackzampolin (approved):

PR #62, opened 2022-03-08
Rename testing.T instances from test to t

mark-rushakoff:
Idiomatic go always names the parameter t.


agouin (approved):
Thanks for this cleanup!

PR #77, opened 2022-06-15
Listen on defined address only

alpe:
## Description
The horcrux server is listening on all network interfaces. With this patch the `p2p-listen` address is taken into account as well.

With `p2p-listen: tcp://10.10.1.1:2222`
` lsof -i -P -n | grep LISTEN`
shows now
`horcrux    7u  IPv6  21097      0t0  TCP *:2222 (LISTEN)`
but it should be
`horcrux    6u  IPv4  21893      0t0  TCP 10.10.1.1:2222 (LISTEN)`

## Checklist

- [ ] I have made sure the upstream branch for this PR is correct
- [ ] I have made sure this PR is ready to merge
- [ ] I have made sure that I have assigned reviewers related to this project

## Changes

- [ ] Added ...
- [ ] Changed ...
- [ ] Removed ...

## Screenshots

...

## Testing

- Open page ...
- Click ...
- Make sure that ...

## Links/References

- ...
- ...
- ...

## Notes

- ...
- ...
- ...

### DIFF
--- a/signer/raft_store.go
+++ b/signer/raft_store.go
@@ -86,11 +86,7 @@ func (s *RaftStore) SetThresholdValidator(thresholdValidator *ThresholdValidator
 
 func (s *RaftStore) init() error {
 	host := p2pURLToRaftAddress(s.RaftBind)
-	_, port, err := net.SplitHostPort(host)
-	if err != nil {
-		return fmt.Errorf("failed to parse local address: %s, %v", host, err)
-	}
-	sock, err := net.Listen("tcp", fmt.Sprintf(":%s", port))
+	sock, err := net.Listen("tcp", host)
 	if err != nil {
 		return err
 	}


agouin (changes requested):
The `p2p-listen` parameter name is a bit confusing because it does not actually listen on the given IP/DNS, only the port.

Raft needs a hostname that is discoverable by the other peers. So this can be an FQDN, for example, which we use for deploying horcrux in kubernetes. This change would break our deployments because it will not be able to listen on a domain name.

From the [migrating doc](https://github.com/strangelove-ventures/horcrux/blob/main/docs/migrating.md):
> The DNS/IP used for this must be reachable by the other peers, i.e. do not use 0.0.0.0 for the hostname

We could add a separate config parameter for the listen IP, but is there a reason you couldn't accomplish this with `nftables` or similar?

alpe:
> The DNS/IP used for this must be reachable by the other peers, i.e. do not use 0.0.0.0 for the hostname

This is configured by the `peers/p2p-addr` fields in the the `config.yaml` file and not touched by this PR.
There is no need to listen on all interfaces for the server side. I have tested this change with a 3 signer setup on a private network and it works all fine.
But `horcrux_test.go:` is failing. I will 👁️ at this asap.

agouin:
> NOTE: The -l or --listen flag lets you set the listen address for the cosigner, which is used for communication between cosigners, Raft and GRPC. The DNS/IP used for this must be reachable by the other peers, i.e. do not use 0.0.0.0 for the hostname.

This is for the `p2p-listen` parameter. From my previous comment:

> Raft needs a hostname that is discoverable by the other peers. So this can be an FQDN, for example, which we use for deploying horcrux in kubernetes. This change would break our deployments because it will not be able to listen on a domain name.

For example, in our deployment in kubernetes,  horcrux cosigner 1 has `p2p-listen` of `tcp://signer-1:2222`. The other 2 cosigners will reference `signer-1` DNS name also in their `peers/p2p-addr` with the same `tcp://signer-1:2222`. Raft needs these addresses to be the same for keeping track of the different nodes, but it is, of course, not actually listening on the DNS name.

> We could add a separate config parameter for the listen IP, but is there a reason you couldn't accomplish this with nftables or similar?

To introduce this change into horcrux, we would need another configuration parameter so that we could inform horcrux of the local interface listen IP in addition to the Raft participant address.

PR #80, opened 2022-06-25
Fix isRunning check

### LINKED ISSUE #79: state set isRunning check not working for me
This [function](https://github.com/strangelove-ventures/horcrux/blob/93e6b7fc4a43ff495514adee5f26bedd7c2defa7/cmd/horcrux/cmd/state.go#L124) does not appear to be working properly.

image

```
root@ip-10-1-1-10:~# ps -ax |pgrep horcrux
root@ip-10-1-1-10:~# horcrux state set 273596
Error: cannot modify state while horcrux is running
```

agouin:
`isRunning` was returning true if the byte output length of `ps ax | pgrep horcrux` was 0, which will always be true because there is a newline in the stdout.

Now it will return true if the number returned by `ps -ax | grep horcrux | wc -l` is greater than 3, which accounts for the 3 running process PIDS that will mention `horcrux`:
- Running horcrux command e.g. `horcrux set state 123`
- `bash -c 'ps -ax | grep horcrux | wc -l'`
- `grep horcrux | wc -l`

Any more than 3 would indicate that another process named horcrux is running, such as the daemon.

Closes #79

### DIFF
--- a/cmd/horcrux/cmd/state.go
+++ b/cmd/horcrux/cmd/state.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path"
 	"strconv"
+	"strings"
 
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
@@ -122,9 +123,14 @@ func setStateCmd() *cobra.Command {
 }
 
 func isRunning() bool {
-	pipe := "ps -ax | pgrep horcrux"
+	pipe := "ps -ax | grep horcrux | wc -l"
 	bz, _ := exec.Command("bash", "-c", pipe).Output()
-	return len(bz) != 0
+	numRunning, err := strconv.ParseUint(strings.TrimSpace(string(bz)), 10, 64)
+	if err != nil {
+		panic(err)
+	}
+	// one is the bash command, one is the grep command, one is this horcrux command. If any more, then horcrux is running as a daemon.
+	return numRunning > 3
 }
 
 func printSignState(ss signer.SignState) {


danbryan:
cool ill give it a test, might be worth looking at something like this libary.

https://github.com/shirou/gopsutil

it looks like it would let you do something like

```
func isRunning(name string) error {
    processes, err := process.Processes()
    if err != nil {
        return err
    }
    for _, p := range processes {
        n, err := p.Name()
        if err != nil {
            return err
        }
        if n == name {
            return true
        }
    }
    return false
}
```

agouin:
> https://github.com/shirou/gopsutil
good find! updated

cmd/horcrux/cmd/state.go
```diff
@@ -5,11 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"os/exec"
 	"path"
+	"path/filepath"
 	"strconv"
 
 	"github.com/mitchellh/go-homedir"
+	"github.com/shirou/gopsutil/process"
```
mark-rushakoff, line 13: gopsutil is on v3 now, so -- if going forward with gopsutil -- the import path should be `github.com/shirou/gopsutil/v3/process`, and go.mod and go.sum will need updated.

mark-rushakoff (commented):
gopsutil seems heavyweight for this usecase. 

Using a lockfile, e.g. with https://pkg.go.dev/github.com/nightlyone/lockfile, would probably be the more straightforward option, if I'm following the use and intent of `isRunning`.

If it's really important to interrogate the OS's processes, https://github.com/mitchellh/go-ps would be one more direct approach.

Alternatively, it doesn't seem that bad to depend on pgrep. I don't see why bash is involved or piping, since pgrep doesn't seem to require anything on stdin. `pgrep --exact horcrux` should exit 0 on match and exit 1 on no match.

cmd/horcrux/cmd/state.go
mark-rushakoff: `pgrep` is probably still fine. `pgrep -d, -x horcrux` would return a list like `14094,23680,80341`, and then you can decide how to process that in Go, whether you just `strings.Count(output, ",")` or actually extract the PIDs and filter away `os.Getpid()`.

That assumes there is no false positive on `pgrep horcrux` and nobody is doing anything funny like renaming the horcrux binary, or using exec and changing `argv[0]`.

The lockfile approach seems a little safer in those regards, as it is much less likely that a rogue file would put something in e.g. `~/.horcrux/horcrux.lock`.

agouin: I see what you mean. The lockfile will be nice too because then you can run multiple horcrux instances out of different config directories and these commands will work properly based on the horcrux runtime state of the applicable config directory.

signer/services.go
```diff
@@ -1,17 +1,25 @@
...
-func WaitAndTerminate(logger tmLog.Logger, services []tmService.Service) {
+func WaitAndTerminate(logger tmLog.Logger, services []tmService.Service, lockFilePath string) {
 	wg := sync.WaitGroup{}
 	wg.Add(1)
+
+	file, err := os.Create(lockFilePath)
+	if err != nil {
+		panic(err)
+	}
+	file.Close()
```
mark-rushakoff, line 20:
```go
err := os.WriteFile(
  lockFilePath,
  []byte(fmt.Sprintf("%d\n", os.Getpid()),
  0600,
)
```

cmd/horcrux/cmd/leader_election.go
```diff
@@ -49,7 +45,7 @@ To choose a specific leader, pass that leader's ID as an argument.
 		}
 
 		var grpcAddresses []string
-		url, err := url.Parse(config.CosignerConfig.P2PListen)
+		url, err := url.Parse(config.Config.CosignerConfig.P2PListen)
```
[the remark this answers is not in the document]
agouin: done.

signer/services.go
agouin: done

signer/services.go
```diff
@@ -1,17 +1,65 @@
...
+
+	trimmed := strings.TrimSpace(string(lockFile))
+
+	pid, err := strconv.ParseInt(trimmed, 10, 64)
+	if err != nil {
+		return fmt.Errorf("unexpected error parsing PID from lock file: %s. manual deletion of lockfile required. %w",
+			lockFilePath, err)
+	}
+
+	_, err = os.FindProcess(int(pid))
```
mark-rushakoff, line 38: > On Unix systems, FindProcess always succeeds and returns a Process for the given pid, regardless of whether the process exists.

Looks like the correct approach is to send it signal 0 to see if the process is alive: https://unix.stackexchange.com/a/169899

signer/services.go
```diff
@@ -1,17 +1,65 @@
...
 	"sync"
 
 	tmLog "github.com/tendermint/tendermint/libs/log"
 	tmOS "github.com/tendermint/tendermint/libs/os"
 	tmService "github.com/tendermint/tendermint/libs/service"
 )
 
-func WaitAndTerminate(logger tmLog.Logger, services []tmService.Service) {
+func RequireNotRunning(lockFilePath string) error {
+	if _, err := os.Stat(lockFilePath); err != nil {
```
mark-rushakoff, line 16: A couple unit tests here would probably go a long way.

cmd/horcrux/cmd/config.go
```diff
@@ -487,59 +433,88 @@ func setChainIDCmd() *cobra.Command {
...
 	}
 	return out
 }
 
+type RuntimeConfig struct {
+	HomeDir    string
+	ConfigFile string
+	KeyFile    string
+	StateDir   string
+	LockFile   string
```
mark-rushakoff, line 502: I think at this point we have a `PidFile` more than a `LockFile`.

signer/services.go
agouin: From what I was reading, it won't have permission to do that unless it was a child process, but I'll try it out

agouin: It worked!

signer/services.go
agouin: added

signer/services.go
```diff
@@ -1,17 +1,83 @@
...
+	if err != nil {
+		return fmt.Errorf(`unclean shutdown detected. PID file exists at %s but PID %d can not be found.
+manual deletion of PID file required. %w`, pidFilePath, pid, err)
+	}
+
+	err = process.Signal(syscall.Signal(0))
+	if err == nil {
+		return fmt.Errorf("horcrux is already running on PID: %d", pid)
+	}
+	if err.Error() == "os: process already finished" {
```
mark-rushakoff, line 47: This should check `errors.Is(err, os.ErrProcessDone)`.

signer/services.go
```diff
@@ -1,17 +1,83 @@
...
+
+func WaitAndTerminate(logger tmLog.Logger, services []tmService.Service, pidFilePath string) {
 	wg := sync.WaitGroup{}
 	wg.Add(1)
+
+	err := os.WriteFile(
+		pidFilePath,
+		[]byte(fmt.Sprintf("%d\n", os.Getpid())),
+		0600,
+	)
```
mark-rushakoff, line 73: What happens if two horcrux processes are started at the same time, neither sees the pid file, and both attempt to write their pid?

This probably needs to switch to `os.OpenFile(pidFilePath, os.O_WRONLY | os.O_CREATE | os.O_EXCL, 0600)` to ensure only one process can create it. And it would be good to get a unit test that uses several goroutines running this function, ensuring that only one succeeds.

signer/services_test.go
```diff
@@ -0,0 +1,92 @@
...
+func TestIsNotRunning(t *testing.T) {
+	homeDir := t.TempDir()
+	pidFilePath := filepath.Join(homeDir, "horcrux.pid")
+
+	err := signer.RequireNotRunning(pidFilePath)
+	require.NoError(t, err)
+}
+
+func getNonExistentPid() (int, error) {
+	maxPidBytes, err := os.ReadFile("/proc/sys/kernel/pid_max")
```
mark-rushakoff, line 43: This may need to call t.Skip based on runtime.GOOS, but that can happen later.

signer/services_test.go
```diff
@@ -0,0 +1,92 @@
...
+	for pid := 1; pid <= int(maxPid); pid++ {
+		process, err := os.FindProcess(pid)
+		if err != nil {
+			continue
+		}
+		err = process.Signal(syscall.Signal(0))
+		if err == nil {
+			continue
+		}
+		if err.Error() == "os: process already finished" {
```
mark-rushakoff, line 60: `errors.Is(err, os.ErrProcessDone)`

agouin: done

signer/services.go
```diff
@@ -1,17 +1,85 @@
...
+	if err != nil {
+		return fmt.Errorf("error reading lock file: %s, %w", pidFilePath, err)
+	}
+
+	pid, err := strconv.ParseInt(strings.TrimSpace(string(lockFile)), 10, 64)
+	if err != nil {
+		return fmt.Errorf("unexpected error parsing PID from lock file: %s. manual deletion of PID file required. %w",
+			pidFilePath, err)
+	}
+
```
mark-rushakoff, line 37: Should we compare `pid` with `os.Getpid()` here? I'm not sure what action that would take -- maybe even panicking would be okay.

mark-rushakoff (approved):
Thanks, this was a bit of a journey, wasn't it?

Please update the skip message and increase the concurrent attempts as mentioned in comments.

agouin:
> Thanks, this was a bit of a journey, wasn't it?

Indeed! Became quite more involved than the `ps` check, but much more reliable, and got to cleanup other things along the way. Thank you.

signer/services_test.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func TestConcurrentStart(t *testing.T) {
+	homeDir := t.TempDir()
+	pidFilePath := filepath.Join(homeDir, "horcrux.pid")
+	var logger tmlog.Logger
+	var services []tmservice.Service
+
+	var wg sync.WaitGroup
+	wg.Add(2)
```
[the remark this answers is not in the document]
agouin: done

signer/services_test.go
```diff
@@ -0,0 +1,165 @@
...
+		if errno == syscall.ESRCH {
+			return pid, nil
+		}
+	}
+	return -1, errors.New("could not find unused PID")
+}
+
+func TestIsRunningNonExistentPid(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		t.Skip()
```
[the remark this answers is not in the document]
agouin: done

signer/services.go
agouin: Added the check. Panic makes sense since it indicates a future coding error, so the stack trace will be helpful there.

PR #83, opened 2022-07-06
Update signing.md

nitronit:
Added some notes to clarify which set-ups are eligible and why.

## Description

This PR ...

## Checklist

- [x] I have made sure the upstream branch for this PR is correct
- [x] I have made sure this PR is ready to merge
- [x] I have made sure that I have assigned reviewers related to this project

## Changes

- [x] Added ...
- [x] Changed ...
- [ ] Removed ...

## Screenshots

...

## Testing

- Open page ...
- Click ...
- Make sure that ...

## Links/References

- ...
- ...
- ...

## Notes

- ...
- ...
- ...

### DIFF
--- a/docs/signing.md
+++ b/docs/signing.md
@@ -19,6 +19,9 @@ _`n: 5`_ and _`t: 3`_ is also a good configuration to enable higher availability
 
 Horcrux is designed with performance in mind, so it will sign and return the full block signature as soon as _`t`_ signer nodes have participated in the block signature.
 
+**N.B**
+Additional to the _Threshold (t)_ >  _Number of shards (n)_/2 requirement steaming from the [Ed25519 Threshold Signatures] layer, the [Raft](https://raft.github.io/)  layer (protocol) requires the number of signer nodes to be an odd number. I.e it means the set up must adhere to the following pattern, (threshold/signer nodes): 2/3, 3/5, 4/7, 5/9 etc.
+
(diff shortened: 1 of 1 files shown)


agouin:
For the number of horcrux cosigners, we don't actually need an odd number. An even number is a somewhat strange configuration because then you don't really get the advantage of availability like you get with the next odd number. E.g. with threshold 3/4 total, you can still only tolerate only one signer being down (same as 2/3), whereas you can tolerate 2 down by just adding one more cosigner and doing a 3/5. However, a 3/4 is more secure than a 2/3 and a 3/5, so it is still a valid configuration.

agouin:
The raft implementation used by horcrux, `hashicorp/raft`, works fine with an even number of nodes also. Similar to how we require `t > n/2` for threshold signing, raft also requires majority nodes for consensus. So with a 3/4 setup, 3 nodes are required for both raft consensus and also threshold signing. It is an odd configuration but not invalid. The 3/4 configuration can be tested in the horcrux test framework in a test such as `TestChainPureHorcrux` by using the following parameters:
```go
	const signersPerValidator = 4
	const threshold = 3
```

nitronit:
I will close this. 

But I stand "firm" that according to the referenced papers you need an odd number of RAFT nodes.

nitronit:
I will close this. 

But I stand "firm" that according to the referenced papers you need an odd number of RAFT nodes.

PR #98, opened 2022-09-16
Threshold small pr

nitronit:
## Description

This PR ...

## Checklist

- [O] I have made sure the upstream branch for this PR is correct
- [O] I have made sure this PR is ready to merge
- [O] I have made sure that I have assigned reviewers related to this project

## Changes

- [ ] Added ...
- [ ] Changed ...
- [ ] Removed ...

## Screenshots

...

## Testing

- Open page ...
- Click ...
- Make sure that ...

## Links/References

- ...
- ...
- ...

## Notes

GLHF @agouin ;)


signer/threshold_signer_soft.go
```diff
@@ -0,0 +1,354 @@
...
+) ThresholdSigner {
+	softSigner := &ThresholdSignerSoft{
+		Key:       key,
+		HrsMeta:   make(map[HRSTKey]HrsMetadata),
+		Total:     total,
+		Threshold: threshold,
+	}
+
+	// cache the public key bytes for signing operations
+	switch ed25519Key := softSigner.Key.PubKey.(type) {
```
agouin, line 44: lets replace this switch case with something like
```golang
ed25519Key := softSigner.Key.PubKey.(tmcryptoed25519.PubKey)
softSigner.PubKeyBytes = make([]byte, len(ed25519Key))
copy(softSigner.PubKeyBytes, ed25519Key[:])
```

As @mark-rushakoff pointed out in the larger PR for a similar issue, the default panic for a failure to cast to `tmcryptoed25519.PubKey` will be better since it will include details about the incorrect type.

nitronit: Should be fixed now.

signer/threshold_signer_soft.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// decrypt share
+	sharePart, err := rsa.DecryptOAEP(sha256.New(), rand.Reader, &softSigner.Key.RSAKey, req.EncryptedSharePart, nil)
+	if err != nil {
+		return err
+	}
+
+	// set slot
+	meta.Peers[req.SourceID-1].Share = sharePart
+	meta.Peers[req.SourceID-1].EphemeralSecretPublicKey = req.SourceEphemeralSecretPublicKey
```
mark-rushakoff, line 345: Ensuring that these statements are going to have an effect, is very subtle.

The HrsMeta map has value types (not pointer types). 

After assigning meta into the map on L334, we are then modifying the local value. Luckily we are indexing into an existing slice, so our value and the value in the map should reference the same slice. Then, we are overwriting the two fields of the value at that index in the slice. Does this write the underlying value in the shared struct? I think so, but it requires careful reading to follow.

It would be much better to have a test demonstrating that this works properly, than requiring the user to follow the code.

signer/threshold_signer_soft.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// Verify the source signature
+	if req.SourceSig == nil {
+		return errors.New("SourceSig field is required")
+	}
+
+	digestMsg := CosignerEphemeralSecretPart{}
+	digestMsg.SourceID = req.SourceID
+	digestMsg.SourceEphemeralSecretPublicKey = req.SourceEphemeralSecretPublicKey
+	digestMsg.EncryptedSharePart = req.EncryptedSharePart
```
mark-rushakoff, line 289: Prefer initialization and assignment together, i.e.
```go
directMsg := CosignerEphemeralSecretPart{
  SourceID: req.SourceID,
  // ...
}
```

mark-rushakoff (commented):
I mostly left stylistic/idiomatic comments, as I am not familiar with the inner workings of threshold signing implementations.

signer/threshold_signer_soft.go
```diff
@@ -0,0 +1,347 @@
...
+// ThresholdSignerSoft implements the interface and signs the message for each local signer.
+// ThresholdSignerSoft is the implementation of a soft sign signer at the local level.
+type ThresholdSignerSoft struct {
+	PubKeyBytes []byte
+	Key         CosignerKey
+	// Total signers
+	Total     uint8
+	Threshold uint8
+	// Height, Round, Step, Timestamp --> metadata
+	HrsMeta map[HRSTKey]HrsMetadata
```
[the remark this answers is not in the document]
nitronit: It "doesnt matter" at the moment since everything is the same package/module. Spoke with Andrew about maybe do a package-refactor later on.

signer/threshold_signer_soft.go
```diff
@@ -0,0 +1,347 @@
...
+		HrsMeta:   make(map[HRSTKey]HrsMetadata),
+		Total:     total,
+		Threshold: threshold,
+	}
+
+	// cache the public key bytes for signing operations.
+	// Ensures casting else it will naturally panic.
+	ed25519Key := softSigner.Key.PubKey.(tmcryptoed25519.PubKey)
+	softSigner.PubKeyBytes = make([]byte, len(ed25519Key))
+	copy(softSigner.PubKeyBytes, ed25519Key[:])
```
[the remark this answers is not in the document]
nitronit: Theoretically yes. But I think that deep copying is more secure as it ensures type casting. Maybe the same goes for softSigner.PubKeyBytes = ed25519Key[:].

signer/threshold_signer_soft.go
nitronit: @mark-rushakoff (and @agouin).
I am having trouble with finding the best way to fix this! 
I 100% agree with @mark-rushakoff  its very subtile and makes the code  almost"impossible" to read and not even a seasoned gopher as you Mark are able to get along. However I am not sure what you mean with "test demonstrating it works". Pure test or a "type casting fix" or a check. 

My solution to fix this is either to use some kind of setter function/method or rewrite and check for softSigner.HrsMeta[hrst].Peers[req.SourceID-1].Share == sharePart or another way to ensure typecasting is maybe to do copy(meta.Peers[req.SourceID-1].Share, sharePart)

Let me know what you think. Its very very very subtile that this is actually legit.

signer/threshold_signer_soft.go
```diff
@@ -0,0 +1,347 @@
...
+		Round:     hrst.Round,
+		Step:      hrst.Step,
+		Signature: sig,
+		SignBytes: req.SignBytes,
+	}, nil, true)
+
+	if err != nil {
+		if _, isSameHRSError := err.(*SameHRSError); !isSameHRSError {
+			return res, err
+		}
```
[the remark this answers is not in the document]
nitronit: Please double check this is correct!

signer/local_cosigner.go
```diff
@@ -11,12 +11,21 @@ import (
...
 
-	tmCryptoEd25519 "github.com/tendermint/tendermint/crypto/ed25519"
-	tmJson "github.com/tendermint/tendermint/libs/json"
+	tmcryptoed25519 "github.com/tendermint/tendermint/crypto/ed25519"
+	tmjson "github.com/tendermint/tendermint/libs/json"
 	"gitlab.com/unit410/edwards25519"
 	tsed25519 "gitlab.com/unit410/threshold-ed25519/pkg"
 )
 
+type LastSignStateStruct struct {
```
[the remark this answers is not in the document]
nitronit: I changed the name to LastSignStateWrapper

nitronit:
I think it can be merged now.

agouin (approved):
Looks great!

PR #102, opened 2022-09-21
go 1.19

### LINKED ISSUE #97: Go version in test and go.mod
Just noticed that go.mod and go test all specifies that go version 1.17 is/should be used. 
However, the latest release is built with 1.16?

Is this something that should be fixed? Should't the test and version run on the same version?

agouin:
Closes #97

### DIFF
--- a/docker/horcrux/Dockerfile
+++ b/docker/horcrux/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM golang:alpine AS build-env
+FROM --platform=$BUILDPLATFORM golang:1.19.1-alpine3.16 AS build-env
 
 ENV PACKAGES make git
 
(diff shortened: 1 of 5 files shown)


DavidNix (approved):

docker/horcrux/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM golang:alpine AS build-env
+FROM --platform=$BUILDPLATFORM golang:1.19.1-alpine3.16 AS build-env
```
mark-rushakoff, line 1: I usually just specify `golang:1.19-alpine` so that we automatically use the latest patch release without having to update this file.

Generally, we would want to adopt go 1.19.<next> as soon as it's available.

Issue #91, opened 2022-08-30
IPV6 not working for elect command


agouin:
Reported by pom dapie in the horcrux telegram channel:
```
horcrux elect
multi:///2001:bc8:***:**::1:3333,2a02:c206:****:****::1:3333,2a01:***:**:****:2::115:3333
Error: rpc error: code = DeadlineExceeded desc = latest balancer error: last connection error: connection error: desc = "transport: Error while dialing dial tcp: address 2a01:***:**:****:2::115:3333: too many colons in address"
```

PR #107, opened 2022-09-29
Fix ipv6 leader election

chillyvee:
## Description

Fixes https://github.com/strangelove-ventures/horcrux/issues/91

## Checklist

- [X] I have made sure the upstream branch for this PR is correct
- [X] I have made sure this PR is ready to merge
- [ ] I have made sure that I have assigned reviewers related to this project

## Changes

Updated string parsing to wrap IPv6 Addresses in [] to distinguish from port

## Testing

Set up horcrux 2 of 3 signing cluster using IPv6 Peer Addresses

horcrux elect 1
horcrux elect 2
horcrux elect 3

## Comments

Purposely did not import a big external dependency to validate IPv4/6 addresses with the assumption that a running cluster likely already has a well formatted and valid configuration file.

### DIFF
--- a/cmd/horcrux/cmd/leader_election.go
+++ b/cmd/horcrux/cmd/leader_election.go
@@ -48,16 +48,24 @@ horcrux elect 2 # elect specific leader`,
 		}
 
 		var grpcAddresses []string
+
+		// Append local host:port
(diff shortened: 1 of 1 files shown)


agouin (approved):
Thank you for this! This also points out that we lack test coverage on leader election, so I will open up a follow-up PR shortly after merging

PR #110, opened 2022-10-15
Leader election tests

agouin:
Moves leader election `multi:///` address construction into a function with unit tests for FQDN, IPv4, and IPv6.
Adds leader election e2e test case `TestLeaderElection2of3`.

Adds ability to fetch current leader via GRPC with `GetLeader`. This is useful for both the tests to make assertions and also in the CLI to log the new leader for more sane command output.

Tidies up code to DRY things up such as `(ts *TestSigner) ExecHorcruxCmd` and `handleNodeJobError` cleanup


cmd/horcrux/cmd/leader_election.go
```diff
@@ -20,6 +20,46 @@ import (
...
+		return fmt.Sprintf("%s:%s", hostname, port), nil
+	}
+}
+
+func leaderElectMultiAddress(cosignerConfig *CosignerConfig) string {
+	var grpcAddresses []string
+
+	// Append local host:port
+	localAddr, err := sanitizeAddress(cosignerConfig.P2PListen)
+	if err == nil {
```
DavidNix, line 50: Is there any danger to swallowing these errors? We never log or return them.

cmd/horcrux/cmd/leader_election_test.go
```diff
@@ -0,0 +1,73 @@
...
+			{
+				P2PAddr: "tcp://10.0.0.2:2222",
+			},
+			{
+				P2PAddr: "tcp://10.0.0.3:2222",
+			},
+		},
+	}
+
+	multiAddress := leaderElectMultiAddress(cfg)
```
DavidNix, line 40: This is opinion.

Testing private functions is generally considered bad practice. See [1](https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice), [2](https://anthonysciamanna.com/2016/02/14/should-private-methods-be-tested.html), [3](https://medium.com/@thrawn01/why-you-should-never-test-private-methods-f822358e010).

I never test private (unexported) methods/functions. Private functionality should be able to change with impunity. A public function clearly signals to the developer that some part of the codebase depends on it, so we have to take greater care of refactoring it. Private functions typically don't signal that importance. 

Additionally, a package with public functions that `main()` functions import surfaces the overall design and components of the system. Vs. hiding them in `main()`.

In this case, I'd recommend creating a new package outside of `cmd` that may have something to do with `leader`.

DavidNix (commented):
Seems like a very useful feature!

cmd/horcrux/cmd/leader_election.go
agouin: good call, errors shouldn't be ignored. updated.

cmd/horcrux/cmd/leader_election_test.go
agouin: good catch, moved these address utils to a `client` package

client/address.go
```diff
@@ -0,0 +1,41 @@
...
+		return "", fmt.Errorf("error splitting host port from parsed URL: %w", err)
+	}
+
+	if strings.Contains(hostname, ":") {
+		// IPv6 Addreses need to be wrapped in brackets
+		return fmt.Sprintf("[%s]:%s", hostname, port), nil
+	} else {
+		return fmt.Sprintf("%s:%s", hostname, port), nil
+	}
+}
```
DavidNix, line 27: This could be simplified with `net.JoinHostPort` which already handles ipv6 for you.

DavidNix: I think instead of splitting then joining host and port, you can simply return `u.Host`. It already removes the scheme and `url.Parse` should already sanitize the address for you.

DavidNix (approved):

Issue #125, opened 2023-02-18
Single config file should work for multiple signers


DavidNix:
This is a loose proposal. 

## Problem to solve
Currently, to configure horcrux, each signer process needs slightly different config. Namely, the listen address and the peers change depending on the signer. You're almost always configuring >1 signer process. So it's error prone to configure multiple signers. Then, if you need to add sentry nodes or other peers, you have to tweak configuration for each signer, which is also error prone. 

This may become more complex when horcrux supports ICS (multiple chains). 

## Proposal

A single config file that can work for any signer. This is similar to the relayer config. In the relayer you set your `chains` and then your `paths`. When you start `rly` you give it the path identifier. Similarly with horcrux, you can configure chains, sentries, peers, etc.. Then feed it a signer identifier. 

For Strangelove, specifically, this simplifies our kubernetes manifests. We can set a signal env var for the signer identifier, for example and minimally tweak the `command` or `args` of a pod template.

PR #130, opened 2023-04-04
Phase 1 of ICS

agouin:
Removes config modification commands such as `horcrux config peers add [peer-nodes]` and `horcrux config peers set-shares` due to confusion. The key needs to be re-sharded every time the peer configuration is updated. Further config cleanup to be handled by #125

`horcrux config set-chain-id` is also removed in preparation of dynamic chain IDs.

Overall config structs are cleaned up/removed throughout to make the constructor parameters explicit.


DavidNix:
I'm taking a look now. Apologies I missed this, you tagged me 3 days ago. My GH notifs weren't setup right for horcrux.

signer/Config.go
```diff
@@ -2,7 +2,12 @@ package signer
 
 import (
```
DavidNix, line 3: Style nit: Go source files are typically lowercase, snake_case. So `Config.go` -> `config.go`.

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
 	"github.com/tendermint/tendermint/crypto"
+	"gopkg.in/yaml.v2"
 )
 
 type NodeConfig struct {
 	Address string
 }
 
+// Config maps to the on-disk JSON format
+type DiskConfig struct {
```
DavidNix, line 29: The name `DiskConfig` makes me think we are configuring a disk. 

I think `signer.Config` is explanatory enough. 

In Go (and other languages), it's good to remember the name of the package contributes to the name as a whole.

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+	}
+	return out
+}
+
+func (c *DiskConfig) CosignerPeers() (out []CosignerParams) {
+	for _, p := range c.CosignerConfig.Peers {
+		out = append(out, CosignerParams{ID: p.ShareID, Address: p.P2PAddr})
+	}
+	return
+}
```
DavidNix, line 57: If it were me, I'd unit test most of these public methods to ensure it produces the output we expect. Maybe they are tested indirectly elsewhere, but still, if someone comes in and changes one of these methods, unit tests will immediately flag the problem area.

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+		panic(err)
+	}
+	return out
+}
+
+func (c *DiskConfig) CosignerPeers() (out []CosignerParams) {
+	for _, p := range c.CosignerConfig.Peers {
+		out = append(out, CosignerParams{ID: p.ShareID, Address: p.P2PAddr})
+	}
+	return
```
DavidNix, line 56: Nit: In Go, many think it's best avoid naked returns even if it shortens LOC. [Issue](https://github.com/golang/go/issues/21291). 

I don't like them because of the indirection. I have to now traverse the function to figure out what, exactly, is being returned.

signer/threshold_validator.go
```diff
@@ -42,36 +44,34 @@ type ThresholdValidator struct {
...
-		cache:    make(map[HRSKey]SignStateConsensus),
+func NewThresholdValidator(
+	config *RuntimeConfig,
+	pubKey crypto.PubKey,
+	signState SignState,
+	threshold int,
+	cosigner Cosigner,
+	peers []Cosigner,
+	raftStore *RaftStore,
+	logger log.Logger,
```
DavidNix, line 56: nit: I'll pretend to be Mark and say the logger should be the first argument to this function. 😆

DavidNix (approved):
Some smallish items but otherwise looks good! Apologies for the very belated review.

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+}
+
+func ChainNodesFromArg(arg string) (out []ChainNode, err error) {
+	for _, n := range strings.Split(arg, ",") {
+		cn := ChainNode{PrivValAddr: n}
+		if err := cn.Validate(); err != nil {
+			return nil, err
+		}
+		out = append(out, cn)
+	}
```
mark-rushakoff, line 240: I think this should collect all the chain nodes and rely on `(ChainNodes).Validate`.

signer/local_cosigner.go
```diff
@@ -109,30 +100,34 @@ func (cosigner *LocalCosigner) SaveLastSignedState(signState SignStateConsensus)
...
+func NewLocalCosigner(
+	config *RuntimeConfig,
+	cosignerKey CosignerKey,
+	signState *SignState,
+	rsaKey rsa.PrivateKey,
+	peers []CosignerPeer,
+	address string,
+	total uint8,
+	threshold uint8,
+) *LocalCosigner {
```
mark-rushakoff, line 112: Eight positional arguments is a bit much. I think the single `LocalCosignerConfig` was a bit better.

signer/threshold_validator.go
mark-rushakoff: Eight positional arguments is a bit much here too.

Putting the logger first (or second if the function accepts `context.Context`) is a simple convention that takes guesswork out of where you should place it. Putting it last can't work with variadic functions.

mark-rushakoff (commented):
I wasn't able to read through everything today, but I can revisit tomorrow. Left a handful of cleanup suggestions.

signer/Config.go
agouin: renamed

signer/Config.go
agouin: unit tests written

signer/Config.go
agouin: definitely, I missed this one. updated.

cmd/horcrux/cmd/cosigner.go
```diff
@@ -45,12 +45,12 @@ func addressCmd() *cobra.Command {
 		SilenceUsage: true,
 		Args:         cobra.RangeArgs(0, 1),
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
-			err = validateCosignerConfig(config.Config)
+			err = config.Config.ValidateCosignerConfig()
 			if err != nil {
 				return
```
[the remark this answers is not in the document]
agouin: updated

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+func (c RuntimeConfig) PrivValStateFile(chainID string) string {
+	return filepath.Join(c.StateDir, fmt.Sprintf("%s_priv_validator_state.json", chainID))
+}
+
+func (c RuntimeConfig) ShareStateFile(chainID string) string {
+	return filepath.Join(c.StateDir, fmt.Sprintf("%s_share_sign_state.json", chainID))
+}
+
+func (c RuntimeConfig) WriteConfigFile() error {
+	return os.WriteFile(c.ConfigFile, c.Config.MustMarshalYaml(), 0644) //nolint
```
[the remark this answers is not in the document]
agouin: good call, 600 is fine. updated

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+}
+
+func (c RuntimeConfig) KeyFileExists(cosigner bool) error {
+	keyFile := c.KeyFilePath(cosigner)
+	if _, err := os.Stat(keyFile); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
agouin: updated to catch more errors and make logs more precise.

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+	// Check that exactly {num-shares}-1 peers are in the peer list, assuming
+	// the remaining peer ID is the ID the local node is configured with.
+	if len(peers) != shares-1 {
+		return fmt.Errorf("incorrect number of peers. expected (%d shares - local node = %d peers)",
+			shares, shares-1)
 	}
 	return nil
 }
 
+func duplicatePeers(peers []CosignerPeerConfig) (duplicates []CosignerPeerConfig) {
```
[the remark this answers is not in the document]
agouin: this is much better, thanks. updated

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+			encountered[peer.ShareID] = peer.P2PAddr
+		} else {
+			duplicates = append(duplicates, CosignerPeerConfig{peer.ShareID, peer.P2PAddr})
+		}
+	}
+	return
+}
+
+func PeersFromFlag(peers string) (out []CosignerPeerConfig, err error) {
+	for _, p := range strings.Split(peers, ",") {
```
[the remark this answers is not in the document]
agouin: updated flag to a StringSlice

signer/local_cosigner.go
agouin: Using a single struct as a param for the constructor of another struct seems like unnecessary overhead to me. The potential zero values in the config struct require extra validate methods, whereas explicit parameters requires setting them all.

These `...Config` types have caused confusion for external contributors, which is the main reason I wanted to remove them.

Any thoughts on improving readability if we keep the `...Config` types?

signer/Config.go
```diff
@@ -11,36 +16,231 @@ import (
...
+	_, err := url.Parse(cn.PrivValAddr)
+	return err
+}
+
+type ChainNodes []ChainNode
+
+func (cns ChainNodes) Validate() error {
+	for _, cn := range cns {
+		if err := cn.Validate(); err != nil {
+			return err
```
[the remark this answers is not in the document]
agouin: #136 for follow up

signer/config.go
```diff
@@ -0,0 +1,281 @@
...
+
+type RuntimeConfig struct {
+	HomeDir    string
+	ConfigFile string
+	StateDir   string
+	PidFile    string
+	Config     Config
+}
+
+func (c RuntimeConfig) KeyFilePath(cosigner bool) string {
```
mark-rushakoff, line 99: I would split this into two different methods, instead of one method accepting a boolean parameter. When I saw `KeyFilePath(true)` earlier, I had no idea what true indicated, nor would I have been able to guess what false meant.

Some discussion on the topic: https://softwareengineering.stackexchange.com/q/147977

mark-rushakoff (approved):

PR #131, opened 2023-04-05
Fix single signer state load bug

agouin:
`LoadFilePVEmptyState` does not load the state file. This was being used unconditionally, so double signing was possible with the single signer mode.

This removes the possibility of double sign by loading the state file on startup for single signers.

### DIFF
--- a/cmd/horcrux/cmd/signer.go
+++ b/cmd/horcrux/cmd/signer.go
@@ -13,6 +13,15 @@ import (
 	"github.com/tendermint/tendermint/types"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/horcrux/cmd/signer.go
```diff
@@ -60,6 +69,8 @@ func StartSignerCmd() *cobra.Command {
 				return err
 			}
 
+			fmt.Println(singleSignerWarning)
```
mark-rushakoff, line 72: This should be using `fmt.Fprintln(cmd.OutOrStdout(), ...)` for more straightforward testing of cobra commands later.

mark-rushakoff (approved):
I didn't read the markdown closely, but the code change LGTM.

DavidNix (approved):

PR #132, opened 2023-04-06
CLI tidy

agouin:
Make CLI methods consistent, remove init funcs. 
Fix test bleed due to `HOME` env var not being respected in tests.


DavidNix (approved):
Global state 🔪

PR #137, opened 2023-04-21
Wait for sign state flush on stop

agouin:
On process termination (and test completion), wait for async sign state writes to complete before shutting down.

Cleanup unnecessary `sync.WaitGroup` in `WaitAndTerminate`, replace with simple chan.

### DIFF
--- a/signer/PvGuard.go
+++ b/signer/PvGuard.go
@@ -8,6 +8,8 @@ import (
 	tm "github.com/tendermint/tendermint/types"
 )
 
+var _ PrivValidator = &PvGuard{}
+
 // PvGuard guards access to an underlying PrivValidator by using mutexes
 // for each of the PrivValidator interface functions
 type PvGuard struct {
@@ -15,23 +17,26 @@ type PvGuard struct {
 	pvMutex       sync.Mutex
 }
 
-// GetPubKey implements types.PrivValidator
+// GetPubKey implements PrivValidator
 func (pv *PvGuard) GetPubKey() (crypto.PubKey, error) {
 	pv.pvMutex.Lock()
 	defer pv.pvMutex.Unlock()
 	return pv.PrivValidator.GetPubKey()
 }
 
-// SignVote implements types.PrivValidator
+// SignVote implements PrivValidator
 func (pv *PvGuard) SignVote(chainID string, vote *tmProto.Vote) error {
 	pv.pvMutex.Lock()
 	defer pv.pvMutex.Unlock()
 	return pv.PrivValidator.SignVote(chainID, vote)
 }
 
-// SignProposal implements types.PrivValidator
+// SignProposal implements PrivValidator
 func (pv *PvGuard) SignProposal(chainID string, proposal *tmProto.Proposal) error {
 	pv.pvMutex.Lock()
 	defer pv.pvMutex.Unlock()
 	return pv.PrivValidator.SignProposal(chainID, proposal)
 }
+
+// SignProposal implements PrivValidator
+func (pv *PvGuard) Stop() {}
--- a/signer/sign_state.go
+++ b/signer/sign_state.go
@@ -107,7 +107,11 @@ func (signState *SignState) GetFromCache(hrs HRSKey, lock *sync.Mutex) (HRSKey, 
 	return latestBlock, nil
 }
 
-func (signState *SignState) Save(ssc SignStateConsensus, lock *sync.Mutex, async bool) error {
+func (signState *SignState) Save(
+	ssc SignStateConsensus,
+	lock *sync.Mutex,
+	asyncWaitGroup *sync.WaitGroup,
+) error {
 	// One lock/unlock for less/equal check and mutation.
 	// Setting nil for lock for getErrorIfLessOrEqual to avoid recursive lock
 	if lock != nil {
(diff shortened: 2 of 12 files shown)


signer/PvGuard.go
```diff
@@ -8,30 +8,35 @@ import (
 	tm "github.com/tendermint/tendermint/types"
 )
 
+var _ PrivValidator = &PvGuard{}
+
 // PvGuard guards access to an underlying PrivValidator by using mutexes
 // for each of the PrivValidator interface functions
 type PvGuard struct {
```
mark-rushakoff, line 15: `SyncPrivValidator` would probably be a bit more idiomatic and clearer name, granting synchronized access to a non-goroutine-safe type. But that can change in another PR, not in scope for this one.

signer/threshold_signer.go
```diff
@@ -24,6 +24,8 @@ type ThresholdSigner interface {
 	Sign(req CosignerSignRequest, m *LastSignStateWrapper) (CosignerSignResponse, error)
 
 	GetID() (int, error)
+
+	Stop()
```
mark-rushakoff, line 28: Worth documenting the expectations around `Stop()` here. It looks like in this case, it is a blocking call that stops any pending goroutines, as opposed to sending an async "stop signal" and requiring a separate `Wait()` call to block on outstanding work.

signer/sign_state.go
```diff
@@ -107,7 +107,11 @@ func (signState *SignState) GetFromCache(hrs HRSKey, lock *sync.Mutex) (HRSKey,
 	return latestBlock, nil
 }
 
-func (signState *SignState) Save(ssc SignStateConsensus, lock *sync.Mutex, async bool) error {
+func (signState *SignState) Save(
```
mark-rushakoff, line 110: This method should be documented. Exported methods accepting a mutex and a wait group are very unusual.

signer/sign_state.go
```diff
@@ -107,7 +107,11 @@ func (signState *SignState) GetFromCache(hrs HRSKey, lock *sync.Mutex) (HRSKey,
...
+	ssc SignStateConsensus,
+	lock *sync.Mutex,
```
mark-rushakoff, line 112: Maybe outside the scope of this PR, but maybe not since you already had to touch all the occurrences of this method:

passing a mutex to a function, just to have it lock and unlock, is an awkward use of a parameter. The caller should manage the lock themselves.

signer/threshold_validator.go
```diff
@@ -75,11 +77,27 @@ func NewThresholdValidator(
...
+}
+
+// Stop safely shuts down the ThresholdValidator.
+func (pv *ThresholdValidator) Stop() {
+	pv.waitForSignStatesToFlushToDisk()
+}
+
+// waitForSignStatesToFlushToDisk waits for any sign states to finish writing to disk.
+func (pv *ThresholdValidator) waitForSignStatesToFlushToDisk() {
+	pv.asyncWaitGroup.Wait()
```
mark-rushakoff, line 94: `asyncWaitGroup` doesn't seem like a specific enough name given this is where the wait occurs. `pendingDiskWG` perhaps?

mark-rushakoff (commented):
Left a few comments, I'll take another read through after this round.

signer/PvGuard.go
agouin: I already renamed it to SingleSignerValidator in the ICS PR, agreed PvGuard is unclear.

signer/threshold_signer.go
agouin: comment added

signer/sign_state.go
agouin: good call, updated.

signer/threshold_validator.go
agouin: renamed

signer/sign_state.go
agouin: documented.

signer/threshold_signer.go
```diff
@@ -24,6 +24,9 @@ type ThresholdSigner interface {
 	Sign(req CosignerSignRequest, m *LastSignStateWrapper) (CosignerSignResponse, error)
 
 	GetID() (int, error)
+
+	// Stop should perform any cleanup work, such as flushing state files to disk, then shut down.
```
mark-rushakoff, line 28: `Stop performs any`...

signer/sign_state.go
```diff
@@ -107,14 +107,14 @@ func (signState *SignState) GetFromCache(hrs HRSKey, lock *sync.Mutex) (HRSKey,
...
 
-func (signState *SignState) Save(ssc SignStateConsensus, lock *sync.Mutex, async bool) error {
-	// One lock/unlock for less/equal check and mutation.
-	// Setting nil for lock for getErrorIfLessOrEqual to avoid recursive lock
-	if lock != nil {
-		lock.Lock()
-		defer lock.Unlock()
-	}
-
+// Save will update the high watermark height/round/step (HRS) if it is greater
```
mark-rushakoff, line 110: `Save updates the`...

mark-rushakoff (approved):
I find this a lot more readable by pulling the mutex out of the Save method.

I commented on two godoc comments, they should be active tense, by convention.

PR #138, opened 2023-04-21
Lint import aliases

agouin:
- Update linter to be more strict around import aliases 
- introduce [revive](https://revive.run/docs)
- handle revive linter errors


signer/config_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestNodes(t *testing.T) {
+func Nodes(t *testing.T) {
```
mark-rushakoff, line 16: Seems wrong -- this will no longer be run as part of `go test`.

signer/grpc_server.go
```diff
@@ -94,8 +94,8 @@ func (rpc *GRPCServer) TransferLeadership(
 }
 
 func (rpc *GRPCServer) GetLeader(
-	ctx context.Context,
-	req *proto.CosignerGRPCGetLeaderRequest,
+	_ context.Context,
+	_ *proto.CosignerGRPCGetLeaderRequest,
```
mark-rushakoff, line 98: When every parameter is `_`, you can omit the names altogether: `GetLeader(context.Context, *proto.CosignerGRPCGetLeaderRequest) (...) { }`.

mark-rushakoff (approved):
Looks straightforward. I think the search and replace caused one test to become not a test anymore, so that should probably be fixed.

DavidNix (approved):

signer/config_test.go
agouin: good catch!

PR #139, opened 2023-04-24
single signer multi chain

agouin:
Prepare SingleSignerValidator for multi-chain-id signing.

Add test case for single signer

### DIFF
--- a/signer/single_signer_validator.go
+++ b/signer/single_signer_validator.go
@@ -1,42 +1,88 @@
 package signer
 
 import (
+	"fmt"
+	"os"
 	"sync"
 
-	tmcrypto "github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/privval"
 	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
-	tm "github.com/tendermint/tendermint/types"
 )
-
-var _ PrivValidator = &SingleSignerValidator{}
 
 // SingleSignerValidator guards access to an underlying PrivValidator by using mutexes
 // for each of the PrivValidator interface functions
 type SingleSignerValidator struct {
-	PrivValidator tm.PrivValidator
-	pvMutex       sync.Mutex
+	config     *RuntimeConfig
+	chainState map[string]*SingleSignerChainState
+}
+
+type SingleSignerChainState struct {
+	filePV  *privval.FilePV
+	pvMutex sync.Mutex
+}
+
+func NewSingleSignerValidator(config *RuntimeConfig) *SingleSignerValidator {
+	return &SingleSignerValidator{
+		config:     config,
+		chainState: make(map[string]*SingleSignerChainState),
+	}
 }
 
 // GetPubKey implements types.PrivValidator
-func (pv *SingleSignerValidator) GetPubKey() (tmcrypto.PubKey, error) {
-	pv.pvMutex.Lock()
-	defer pv.pvMutex.Unlock()
-	return pv.PrivValidator.GetPubKey()
+func (pv *SingleSignerValidator) GetPubKey() (crypto.PubKey, error) {
+	keyFile := pv.config.KeyFilePathSingleSigner()
+	filePV := privval.LoadFilePVEmptyState(keyFile, "")
+	return filePV.GetPubKey()
 }
 
 // SignVote implements types.PrivValidator
 func (pv *SingleSignerValidator) SignVote(chainID string, vote *tmproto.Vote) error {
-	pv.pvMutex.Lock()
-	defer pv.pvMutex.Unlock()
-	return pv.PrivValidator.SignVote(chainID, vote)
+	chainState, err := pv.loadChainStateIfNecessary(chainID)
+	if err != nil {
+		return err
+	}
+	chainState.pvMutex.Lock()
+	defer chainState.pvMutex.Unlock()
+	return chainState.filePV.SignVote(chainID, vote)
 }
 
 // SignProposal implements types.PrivValidator
 func (pv *SingleSignerValidator) SignProposal(chainID string, proposal *tmproto.Proposal) error {
-	pv.pvMutex.Lock()
-	defer pv.pvMutex.Unlock()
-	return pv.PrivValidator.SignProposal(chainID, proposal)
+	chainState, err := pv.loadChainStateIfNecessary(chainID)
+	if err != nil {
+		return err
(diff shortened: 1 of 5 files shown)


signer/single_signer_validator.go
```diff
@@ -1,42 +1,90 @@
...
 
 // GetPubKey implements types.PrivValidator
-func (pv *SingleSignerValidator) GetPubKey() (tmcrypto.PubKey, error) {
-	pv.pvMutex.Lock()
-	defer pv.pvMutex.Unlock()
-	return pv.PrivValidator.GetPubKey()
+func (pv *SingleSignerValidator) GetPubKey() (crypto.PubKey, error) {
+	keyFile := pv.config.KeyFilePathSingleSigner()
+	filePV := privval.LoadFilePVEmptyState(keyFile, "")
+	return filePV.GetPubKey()
```
DavidNix, line 38: Is there a reason we want to load all the information on each method invocation?

Or can we load it once (such as close to `main()`) and then populate a field in `SingleSignerValidator`?

signer/single_signer_validator.go
```diff
@@ -1,42 +1,90 @@
...
+	} else {
+		filePV = privval.LoadFilePV(keyFile, stateFile)
+	}
+
+	chainState := &SingleSignerChainState{
+		filePV: filePV,
+	}
+	pv.chainState[chainID] = chainState
+
+	return chainState, nil
```
DavidNix, line 87: Same thing here. I may be missing why we need to load files off disk and make `os` calls like `os.Stat` on each method invocation. It seems like this loading of config could be done once then you populate the struct with the values it needs. These values, then, would never be mutated, only read. And then you can eliminate the mutex. 

Hiding `os` calls inside these methods are a bit surprising to me. I wouldn't expect it unless the struct took in something like an [fs.FS](https://pkg.go.dev/io/fs#FS).

signer/single_signer_validator.go
agouin: `GetPubKey` is called once by the sentry after initializing the connection. The sentry uses this to determine if the remote signer is a validator or not. I thought about checking `len(chainState) > 0` then using the pub key from the first one (pub key from all will be the same), but that would require doing this anyways when the chainID does not yet exist in the map (which will always be the case since GetPubKey is called before doing any specific chainID operations)

signer/single_signer_validator.go
agouin: caching is implemented via L64-L66 and L85. disk reads will only happen on the first invocation.

agouin: The mutex is used as a safeguard so that only one sign operation can happen at a time, not for file operations. But that makes me realize, we need a separate mutex for the map writes.

signer/single_signer_validator.go
agouin: I updated it to load the pub key during construction

signer/single_signer_validator.go
DavidNix: I see. So I the `chainID` is only known at runtime? Or Is there a way to know all the potential chain ids when the process starts? 

My ideal is to hydrate all the `chainState`s during process initialization, unless there is some advantage to lazy loading as implemented here. 

> 1 mutex is typically a code smell (and the cause of insidious deadlocks). You could opt for a [sync.Map](https://pkg.go.dev/sync#Map)

agouin: Currently, horcrux is has a `chain-id` in the yaml config, and uses that value for signing votes and proposals. This only works if horcrux is signing for a single chain, however. In #122 , the config value is removed/ignored and the chain ID from the vote/proposal sign req is used instead. This means we need to load the state file upon the first vote/proposal sign req for each new chain ID.

With the order of how these mutexes are used, there is no chance of deadlock. But thanks for pointing out the sync.Map. One of it's optimized use cases is `(1) when the entry for a given key is only ever written once but read many times, as in caches that only grow`, which is the case for the `chainState` map.

DavidNix: > the config value is removed/ignored and the chain ID from the vote/proposal sign req is used instead.

This was the missing piece for me, thanks! In that case, we have load dynamically at runtime. I still bet there is a patten to remove the os calls, but that can be a refactor in the future.

signer/single_signer_validator.go
```diff
@@ -1,42 +1,116 @@
...
+	keyFile := pv.config.KeyFilePathSingleSigner()
+	stateFile := pv.config.PrivValStateFile(chainID)
+	var filePV *tmprivval.FilePV
+	if _, err := os.Stat(stateFile); err != nil {
+		if !os.IsNotExist(err) {
+			panic(fmt.Errorf("failed to load state file (%s) - %w", stateFile, err))
+		}
+		// The only scenario in which we want to initialize a new state file
+		// is when the state file does not exist.
+		filePV = tmprivval.LoadFilePVEmptyState(keyFile, stateFile)
```
DavidNix, line 103: Ah, `LoadFilePVEmptyState` appears to create a new file on disk. 

If that's the case:

Removing the os calls from these methods would be difficult. 

That function name does not make that side effect obvious. Perhaps adding to the comments here (which are already good) to explicitly state we are creating a new file on disk.

signer/single_signer_validator.go
```diff
@@ -1,42 +1,116 @@
...
-	defer pv.pvMutex.Unlock()
-	return pv.PrivValidator.SignVote(chainID, vote)
+	chainState, err := pv.loadChainStateIfNecessary(chainID)
+	if err != nil {
+		return err
+	}
+	chainState.pvMutex.Lock()
+	defer chainState.pvMutex.Unlock()
+	return chainState.filePV.SignVote(chainID, vote)
 }
```
DavidNix, line 75: I swear I asked this already but cannot find the GH comment anywhere. Apologies about the double ask.

What does this mutex protect? I see it's locking per chainID, which is fine.

If it's to protect against a double sign, I don't see how it would work. That's the caller's responsibility. 

The caller could do:

```go
vote := getVoteSomehow()
for {
   validator.SignVote("cosmoshub-4", vote)
}
```

The same applies for `SignProposal`.

DavidNix (approved):

signer/single_signer_validator.go
agouin: Updated the comment

signer/single_signer_validator.go
DavidNix: Thanks for that explanation. That last comment would be great in go doc comments.

PR #140, opened 2023-04-25
Make threshold validator chain ID agnostic

agouin:
Next phase of #122 

Prepare ThresholdValidator for multi-chain-id signing.


DavidNix (commented):
Definitely making good progress!

One overall question: Is this new code backwards compatible with the old config where you specify the chain id?

agouin:
> One overall question: Is this new code backwards compatible with the old config where you specify the chain id?

In this PR, it does not change anything about config or CLI commands to keep it as slim as possible. In the continuation of this in #122, it removes `chain-id` from the config, the `config init` cmd, and requires `chain-id` as an arg for the cmds that interact with the state files. If an older config that still has `chain-id` is used, it will be ignored. So yes, it is backwards compatible.

DavidNix (approved):
thanks 🙏

PR #141, opened 2023-04-26
go 1.20

### LINKED ISSUE #136: Update to `go` 1.20
Update go.mod, docker images, and github workflows to go 1.20

Good opportunities for enriching errors with `errors.Join`:
https://github.com/strangelove-ventures/horcrux/pull/130#discussion_r1169247422
https://github.com/strangelove-ventures/horcrux/pull/130#discussion_r1169261793

agouin:
Bump to go 1.20 and implement errors.Join suggestions

Resolves #136


signer/services_test.go
```diff
@@ -17,29 +17,37 @@ import (
...
+}
+
+func TestIsRunning(t *testing.T) {
+	homeDir := t.TempDir()
+	pidFilePath := filepath.Join(homeDir, "horcrux.pid")
+
+	err := fork.Fork("child", pidFilePath)
+	require.NoError(t, err)
+
+	time.Sleep(time.Second * 1)
```
mark-rushakoff, line 44: Since this is a relatively long sleep, this test should use t.Parallel.

signer/services_test.go
```diff
@@ -17,29 +17,37 @@ import (
...
 		0600,
 	)
-	require.NoError(t, err, "error writing pid file")
+}
+
+func TestIsRunning(t *testing.T) {
+	homeDir := t.TempDir()
+	pidFilePath := filepath.Join(homeDir, "horcrux.pid")
+
+	err := fork.Fork("child", pidFilePath)
```
mark-rushakoff, line 41: Comment what is going on with this fork package/function. It looks like this might only be used in test? It's an unfamiliar package.

mark-rushakoff (approved):
FYI errors.Join handles nil elements in its arguments correctly (by ignoring them).

The nil checks are fine to keep , since that avoids allocating / growing a slice, but it may be about as readable with an unconditional append. I'm fine either way since the erroring path is very unlikely to be a hotspot.

.github/workflows/release.yml
```diff
@@ -14,13 +14,13 @@ jobs:
...
-        uses: actions/checkout@v2.3.4
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-go@v2
+      - uses: actions/setup-go@v4
         with:
-          go-version: '1.19'
+          go-version: '1.20.3'
```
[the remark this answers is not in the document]
agouin: We learned in the relayer that `'1.20'` for `actions/setup-go` will actually use go 1.20 (the first) instead of the latest (since go does not use `1.20.0` for the first 1.20 release for some reason).

Checking their docs now though, it looks like we can use `^1.20.0` to use the latest 1.20.x

signer/services_test.go
agouin: comment added

signer/services_test.go
agouin: I removed this sleep and replaced it with a waitForFileToExist tick/timeout, with a 1 second timeout.

PR #144, opened 2023-05-02
key assignment feature

### LINKED ISSUE #135: Replicated Security - support key assignment feature
For Replicated Security there are 2 ways to sign blocks. The first and simplest one is to use the same `priv_validator_key.json` for provider and consumer chains. The problem with this solution is that it increases the insecurity of the Hub, and more in the future as RS scale.

As describe in this [blog post](https://forum.cosmos.network/t/swiss-booklet-launch-neutron-on-replicated-security/10396#how-is-the-hub-protected-from-malicious-consumer-chains-11):

> validators should use the [key assignment feature](https://github.com/cosmos/interchain-security/blob/main/docs/docs/features/key-assignment.md) so that they can keep their provider chain and consumer chain keys separated.

Key assignment feature allows validators to use different keys for consumer and provider chains by broadcasting to the network which consensus key they will be using to validate a consumer chain (`providerd tx provider assign-consensus-key <consumer_pub_key>`)

For the launch of first consumer chain (Neutron) its still manageable to use the same key, but I think in the future it could be great that Horcrux supports RS key assignment!

agouin:
Splits up keys to require separate key file for each chain ID

Changes `--keyfile` flag to `--key-dir`, as there can be multiple keys. Keys must be named `{chain_id}_share.json` for threshold signer mode, or `{chain_id}_priv_validator_key.json` for single signer mode.

Breaks up cosigner RSA keys into separate `cosigner.json` key file (RSA config is one key per cosigner, but ed25519 config is potentially many key shards per signer). Sharding a priv-validator key is still done with `horcrux create-shares`, but RSA keys must be created separately with `horcrux create-shares-rsa`

Resolves #135


cmd/horcrux/cmd/key2shares.go
```diff
@@ -51,39 +85,62 @@ func createCosignerSharesCmd() *cobra.Command {
...
+				dir, err := createCosignerDirectoryIfNecessary(c.ID)
+				if err != nil {
 					return err
 				}
-				fmt.Printf("Created Share %d\n", c.ID)
+				filename := filepath.Join(dir, fmt.Sprintf("%s_share.json", chainID))
+				if err = signer.WriteCosignerShareFile(c, filename); err != nil {
+					return err
+				}
+				fmt.Printf("Created Ed25519 Share %s\n", filename)
```
mark-rushakoff, line 96: Prefer `fmt.Fprintf(cmd.OutOrStdout(), "...")`.

cmd/horcrux/cmd/key2shares.go
```diff
@@ -51,39 +85,62 @@ func createCosignerSharesCmd() *cobra.Command {
...
-	if t > n {
-		return fmt.Errorf("threshold cannot be greater than total shares, got [threshold](%d) > [shares](%d)", t, n)
-	}
-	if t <= n/2 {
-		return fmt.Errorf("threshold must be greater than total shares "+
-			"divided by 2, got [threshold](%d) <= [shares](%d) / 2", t, n)
+// CreateCosignerSharesCmd is a cobra command for creating cosigner shares from a priv validator
+func createCosignerRSASharesCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "create-rsa-shares [shares]",
```
mark-rushakoff, line 107: No brackets.

cmd/horcrux/cmd/key2shares.go
```diff
@@ -51,39 +85,62 @@ func createCosignerSharesCmd() *cobra.Command {
...
+		Aliases: []string{"shard", "shares"},
+		Args:    cobra.ExactArgs(1),
+		Short:   "Create  cosigner shares",
```
mark-rushakoff, line 110: Double-space in string.

cmd/horcrux/cmd/key2shares.go
```diff
@@ -51,39 +85,62 @@ func createCosignerSharesCmd() *cobra.Command {
...
+			for _, c := range csKeys {
+				dir, err := createCosignerDirectoryIfNecessary(c.ID)
+				if err != nil {
+					return err
+				}
+				filename := filepath.Join(dir, "rsa_keys.json")
+				if err = signer.WriteCosignerShareRSAFile(c, filename); err != nil {
+					return err
+				}
+				fmt.Printf("Created RSA Share %s\n", filename)
```
mark-rushakoff, line 140: Print to `cmd.OutOrStdout()`.

cmd/horcrux/cmd/key2shares_test.go
```diff
@@ -25,41 +28,91 @@ func TestKey2Shares(t *testing.T) {
...
-			args:      []string{privValidatorKeyFile, "two", "three"},
+			args:      []string{testChainID, privValidatorKeyFile, "two", "three"},
+			expectErr: true,
+		},
+	}
+
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+
+	require.NoError(t, os.Chdir(tmp))
```
mark-rushakoff, line 64: It doesn't look like changing the directory is necessary, unless there is something dependent about the working directory for the command. 

It's unusual. If it is required, the change back to the original directory should be in a `t.Cleanup` call so that there isn't a risk of the call being missed during an intermediate failure.

cmd/horcrux/cmd/key2shares_test.go
```diff
@@ -25,41 +28,91 @@ func TestKey2Shares(t *testing.T) {
...
+			expectErr: false,
+		},
+		{
+			name:      "invalid shares",
+			args:      []string{"0"},
 			expectErr: true,
 		},
 	}
 
+	cwd, err := os.Getwd()
```
mark-rushakoff, line 105: Same for this test.

cmd/horcrux/cmd/signer.go
```diff
@@ -55,34 +54,19 @@ func startSignerCmd() *cobra.Command {
...
-			)
-
-			_, err = config.KeyFileExistsSingleSigner()
-			if err != nil {
-				return err
-			}
-
-			logger.Info("CometBFT Validator", "mode", "single-signer",
-				"priv-key", config.Config.PrivValKeyFile, "priv-state-dir", config.StateDir)
+			logger := cometlog.NewTMLogger(cometlog.NewSyncWriter(os.Stdout)).With("module", "validator")
```
mark-rushakoff, line 57: This should use `cmd.OutOrStdout()` instead of `os.Stdout` -- I think I may have missed at least one earlier occurrence of this.

signer/cosigner_key.go
```diff
@@ -113,13 +160,12 @@ func (cosignerKey *CosignerKey) UnmarshalJSON(data []byte) error {
 	}
 
 	cosignerKey.RSAKey = *privateKey
-	cosignerKey.PubKey = pubkey
 	return nil
 }
 
 // LoadCosignerKey loads a CosignerKey from file.
```
mark-rushakoff, line 166: Outdated godoc.

signer/threshold_validator_test.go
```diff
@@ -173,42 +181,88 @@ func TestThresholdValidator3of3(t *testing.T) {
 	privKeyBytes := privateKey[:]
 	secretShares := tsed25519.DealShares(tsed25519.ExpandSecret(privKeyBytes[:32]), threshold, total)
 
+	tmpDir := t.TempDir()
+
+	var cosigner1, cosigner2, cosigner3 Cosigner
```
mark-rushakoff, line 186: If there is a way to pull most of this code into a loop, that would help readability of this test.

cmd/horcrux/cmd/migrate_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMigrateV2toV3(t *testing.T) {
+	v2ConfigBz, err := os.ReadFile("./fixtures/config-v2.yaml")
```
mark-rushakoff, line 14: Shouldn't happen in this PR, but typically this would be reading from a `testdata` directory, as that name has special meaning to the Go toolchain.

Furthermore, unless this actually _needs_ to interact with disk, I prefer to lean on `//go:embed` here, so that if we ever need a test binary, it isn't dependent on the local filesystem. (It works with `go test` because that commands sets the working directory to the package being tested.)

mark-rushakoff (approved):
This was a fairly large PR, so I didn't read it as closely as I would for a smaller PR. Noticed a few small style things, but nothing seems wrong as far as I can see.

cmd/horcrux/cmd/key2shares_test.go
agouin: These create shares commands output the files to the cwd. Would be nice to have a flag to change the output dir so that the tests can avoid this. Will modify

cmd/horcrux/cmd/migrate_test.go
DavidNix: +1 for `testdata` dir and `go:embed`

cmd/horcrux/cmd/key2shares.go
```diff
@@ -17,32 +17,66 @@ package cmd
...
+
 // CreateCosignerSharesCmd is a cobra command for creating cosigner shares from a priv validator
-func createCosignerSharesCmd() *cobra.Command {
+func createCosignerEd25519SharesCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "create-shares [priv_validator.json] [threshold] [shares]",
+		Use:     "create-ed25519-shares [chain-id] [priv_validator.json] [threshold] [shares]",
 		Aliases: []string{"shard", "shares"},
-		Args:    validateCreateCosignerShares,
+		Args:    cobra.ExactArgs(4),
```
DavidNix, line 51: \> 1 or 2 positional args gets confusing, especially since at the end of the day they are all strings. 

Consider using required flags instead. I think that improves U/X. Personally, when calling this command I'm always afraid I mix up the threshold and the shares. Flags make it explicit. 

```
--chain-id
--key-file
--threshold
--shares
```

DavidNix:
I'm taking a look. I'll have some comments. Give me a little bit because this is a large PR.

cmd/horcrux/cmd/key2shares.go
DavidNix: As a bonus, with cobra, you don't have to do the string conversion if they are flags.

signer/threshold_validator.go
```diff
@@ -351,9 +391,20 @@ func (pv *ThresholdValidator) getExistingBlockSignature(chainID string, block *B
...
+
+	cs, ok := pv.chainState.Load(chainID)
+	if !ok {
+		return nil, stamp, fmt.Errorf("failed to load chain state for %s", chainID)
+	}
+
+	css, ok := cs.(ChainSignState)
+	if !ok {
+		return nil, stamp, fmt.Errorf("expected: (ChainSignState), actual: (%T)", cs)
+	}
```
DavidNix, line 403: This bit of typecasting is repeated a lot. It may clean things up to abstract to our own type or functions. Go should have a generic sync map by now. :( 

Also, I think it's ok to panic here instead of return an error. If anything other than the type we want is present, something is seriously wrong. Test coverage should immediately flag that panic.

DavidNix (commented):
I ❤️ the design here as well as giving users a migrate command.

I forgot where this was suggested (or maybe I dreamed it): In the future, splitting a PR like this into two, key dir refactor and migrate command, would probably help with code review. 

I missed this, but we save keys into 1 dir and name files with a chain-id suffix? Or are there subdirs also named after the chain-id?

cmd/horcrux/cmd/key2shares.go
DavidNix: Can be a future PR if you agree.

cmd/horcrux/cmd/cosigner.go
```diff
@@ -117,21 +117,26 @@ func startCosignerCmd() *cobra.Command {
 				return err
 			}
 
-			var (
-				// services to stop on shutdown
-				services []cometservice.Service
-				logger   = cometlog.NewTMLogger(cometlog.NewSyncWriter(os.Stdout)).With("module", "validator")
+			logger := cometlog.NewTMLogger(cometlog.NewSyncWriter(os.Stdout)).With("module", "validator")
```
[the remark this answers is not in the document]
agouin: It would be great to swap out logging to slog or zap in horcrux. I created #146 as follow up

cmd/horcrux/cmd/fixtures/config-migrated.yaml
```diff
@@ -0,0 +1,14 @@
...
+  peers:
+  - share-id: 2
+    p2p-addr: tcp://127.0.0.1:2223
+  - share-id: 3
+    p2p-addr: tcp://127.0.0.1:2224
+  rpc-timeout: 1000ms
+chain-nodes:
+- priv-val-addr: tcp://127.0.0.1:1234
+- priv-val-addr: tcp://127.0.0.1:2345
+- priv-val-addr: tcp://127.0.0.1:3456
```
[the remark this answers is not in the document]
agouin: Agreed, but I don't think it's necessary to break the entire config for v3 :smile: . I suppose we could add it to the migrate command, however.

mark-rushakoff:
I’m switching to slog for all my new work, since it is on track to be part
of the standard library.

cmd/horcrux/cmd/fixtures/config-migrated.yaml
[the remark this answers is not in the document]
DavidNix: 👍

cmd/horcrux/cmd/cosigner.go
```diff
@@ -37,18 +37,18 @@ type AddressCmdOutput struct {
 
 func addressCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:          "address [bech32]",
+		Use:          "address [chain-id] [bech32]",
```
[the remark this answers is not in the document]
agouin: thanks for pointing that out, all catches updated

cmd/horcrux/cmd/key2shares_test.go
agouin: Flag added, test updated

signer/cosigner_key.go
agouin: updated

signer/threshold_validator_test.go
agouin: refactored this file to clean up test into a single test body with dynamic test cases for the various threshold/shares combos.

cmd/horcrux/cmd/migrate_test.go
agouin: definitely nice cleanup. migrated

cmd/horcrux/cmd/key2shares.go
agouin: good UX gain. updated to use the flags

signer/threshold_validator.go
agouin: refactored out into a mustLoadChainState method

agouin:
> I missed this, but we save keys into 1 dir and name files with a chain-id suffix?

Sharding will create a dir for each cosigner. The files can be copied directly to the relative cosigners without renaming any files now.

```
$ ls
priv_validator_key.json

$ horcrux create-ed25519-shares --chain-id cosmoshub-4 --key-file priv_validator_key.json --threshold 2 --shares 3
Created Ed25519 Share cosigner_1/cosmoshub-4_share.json
Created Ed25519 Share cosigner_2/cosmoshub-4_share.json
Created Ed25519 Share cosigner_3/cosmoshub-4_share.json

$ horcrux create-rsa-shares --shares 3
Created RSA Share cosigner_1/rsa_keys.json
Created RSA Share cosigner_2/rsa_keys.json
Created RSA Share cosigner_3/rsa_keys.json

$ ls -R
.:
cosigner_1  cosigner_2  cosigner_3 priv_validator_key.json

./cosigner_1:
cosmoshub-4_share.json  rsa_keys.json

./cosigner_2:
cosmoshub-4_share.json  rsa_keys.json

./cosigner_3:
cosmoshub-4_share.json  rsa_keys.json
```

DavidNix (approved):

PR #147, opened 2023-05-03
One config file to rule them all and in cryptography bind them

### LINKED ISSUE #125: Single config file should work for multiple signers

agouin:
Merges `cosigner.p2p-listen` into single list of cosigners. The cosigner shard ID (in both `rsa_keys.json` and `{chain_id}_share.json`, which must match), will be used to determine which address should be used for the current cosigner.

This enables a single config file to be shared across all cosigners, and also means `horcrux config init` can be used once to generate that shared config.

Renames some config properties for clarity, updating `horcrux config migrate` to provide smooth migration path from v2 config.

Adds `signMode` to config. This will be used for `horcrux start`, which replaces `horcrux signer start` and `horcrux cosigner start`.

Consistent naming convention: horcrux `threshold` mode consists of `cosigners`. Cosigners have key `shards`. Those shards have `ID`s.

### Before:
Cosigner 1
```
chain-id: test
cosigner:
  threshold: 2
  shares: 3
  p2p-listen: tcp://127.0.0.1:2222
  peers:
  - share-id: 2
    p2p-addr: tcp://127.0.0.1:2223
  - share-id: 3
    p2p-addr: tcp://127.0.0.1:2224
  rpc-timeout: 1000ms
chain-nodes:
- priv-val-addr: tcp://127.0.0.1:1234
- priv-val-addr: tcp://127.0.0.1:2345
- priv-val-addr: tcp://127.0.0.1:3456
```
Cosigner 2
```
chain-id: test
cosigner:
  threshold: 2
  shares: 3
  p2p-listen: tcp://127.0.0.1:2223
  peers:
  - share-id: 1
    p2p-addr: tcp://127.0.0.1:2222
  - share-id: 3
    p2p-addr: tcp://127.0.0.1:2224
  rpc-timeout: 1000ms
chain-nodes:
- priv-val-addr: tcp://127.0.0.1:1234
- priv-val-addr: tcp://127.0.0.1:2345
- priv-val-addr: tcp://127.0.0.1:3456
```
Cosigner 3
```
chain-id: test
cosigner:
  threshold: 2
  shares: 3
  p2p-listen: tcp://127.0.0.1:2224
  peers:
  - share-id: 1
    p2p-addr: tcp://127.0.0.1:2222
  - share-id: 2
    p2p-addr: tcp://127.0.0.1:2223
  rpc-timeout: 1000ms
chain-nodes:
- priv-val-addr: tcp://127.0.0.1:1234
- priv-val-addr: tcp://127.0.0.1:2345
- priv-val-addr: tcp://127.0.0.1:3456
```

### After
All Cosigners
```
signMode: threshold
thresholdMode:
  threshold: 2
  cosigners:
  - shardID: 3
    p2pAddr: tcp://127.0.0.1:2224
  - shardID: 1
    p2pAddr: tcp://127.0.0.1:2222
  - shardID: 2
    p2pAddr: tcp://127.0.0.1:2223
  grpcTimeout: 1000ms
  raftTimeout: 1000ms
chainNodes:
- privValAddr: tcp://127.0.0.1:1234
- privValAddr: tcp://127.0.0.1:2345
- privValAddr: tcp://127.0.0.1:3456
```

Resolves #125


signer/config_test.go
```diff
@@ -523,7 +543,7 @@ func TestCosignerPeersConfigValidate(t *testing.T) {
 					P2PAddr: "tcp://127.0.0.1:2223",
 				},
 			},
-			shares:    3,
+			shares:    2,
```
DavidNix, line 546: Does the user need to specify the # of shares? Could we use `len(peers)` instead?

signer/local_cosigner_test.go
```diff
@@ -35,15 +35,16 @@ func TestLocalCosignerGetID(t *testing.T) {
 
 	cosigner := NewLocalCosigner(
 		&RuntimeConfig{},
-		key.ID,
-		*rsaKey,
+		CosignerKeyRSA{
+			ID:     key.ID,
+			RSAKey: *rsaKey,
+		},
```
DavidNix, line 41: Possibly for a future PR, and a question I've been meaning to ask is why we need the RSA keys.

I found reference to the RSA keys horcrux/docs/signing.md which illuminates some. 

But I think we could expand more explaining why we need them, how to create, and how to manage.

signer/threshold_validator.go
```diff
@@ -222,7 +222,7 @@ func (pv *ThresholdValidator) waitForPeerEphemeralShares(
 	peer Cosigner,
 	hrst HRSTKey,
 	wg *sync.WaitGroup,
-	encryptedEphemeralSharesThresholdMap *map[Cosigner][]CosignerEphemeralSecretPart,
+	encryptedEphemeralSharesThresholdMap map[Cosigner][]CosignerEphemeralSecretPart,
```
DavidNix, line 225: 👏 Pointers to maps or slices are very uncommon.

signer/threshold_validator.go
```diff
@@ -253,7 +253,7 @@ func (pv *ThresholdValidator) waitForPeerSetEphemeralSharesAndSign(
 	ourID int,
 	peer Cosigner,
 	hrst HRSTKey,
-	encryptedEphemeralSharesThresholdMap *map[Cosigner][]CosignerEphemeralSecretPart,
+	encryptedEphemeralSharesThresholdMap map[Cosigner][]CosignerEphemeralSecretPart,
```
DavidNix, line 256: nit: It's rare to see long variable names in Go. `encryptedShares` is probably descriptive enough. You can quickly see the type for more info. I can't tell here, but `map[Cosigner][]CosignerEphemeralSecretPart` could be more readable as its own type or type alias.

signer/threshold_validator.go
```diff
@@ -263,11 +263,12 @@ func (pv *ThresholdValidator) waitForPeerSetEphemeralSharesAndSign(
...
+
+	for _, encryptedSecrets := range encryptedEphemeralSharesThresholdMap {
+		for _, ephemeralSecretPart := range encryptedSecrets {
 			// if share is intended for peer, check to make sure source peer is included in threshold
 			if ephemeralSecretPart.DestinationID == peer.GetID() {
-				for thresholdPeer := range *encryptedEphemeralSharesThresholdMap {
+				for thresholdPeer := range encryptedEphemeralSharesThresholdMap {
 					if thresholdPeer.GetID() == ephemeralSecretPart.SourceID {
 						// source peer is included in threshold signature, include in sharing
 						peerEphemeralSecretParts = append(peerEphemeralSecretParts, ephemeralSecretPart)
```
DavidNix, line 274: No change because this is existing code that is already working. But there are 3 for loops (2 nested) and 2 nested if statements. I bet there's a refactor that could make this more readable in the future.

signer/threshold_validator.go
```diff
@@ -263,11 +263,12 @@ func (pv *ThresholdValidator) waitForPeerSetEphemeralSharesAndSign(
...
-	for _, EncryptedSecrets := range *encryptedEphemeralSharesThresholdMap {
-		for _, ephemeralSecretPart := range EncryptedSecrets {
+
+	for _, encryptedSecrets := range encryptedEphemeralSharesThresholdMap {
+		for _, ephemeralSecretPart := range encryptedSecrets {
 			// if share is intended for peer, check to make sure source peer is included in threshold
 			if ephemeralSecretPart.DestinationID == peer.GetID() {
-				for thresholdPeer := range *encryptedEphemeralSharesThresholdMap {
+				for thresholdPeer := range encryptedEphemeralSharesThresholdMap {
 					if thresholdPeer.GetID() == ephemeralSecretPart.SourceID {
```
DavidNix, line 272: Re naming consistency, indicates peer id.

DavidNix (commented):
Sorry, I left a lot comments. Happy to discuss outside of GH. 

My main concern is naming consistency. Share id, peer id, cosigner id, etc. It would be great to be consistent with this id. From this PR, `peer id` seems best to me.

cmd/horcrux/cmd/config_test.go
```diff
@@ -23,9 +23,8 @@ func TestConfigInitCmd(t *testing.T) {
 			args: []string{
 				"tcp://10.168.0.1:1234",
 				"-c",
-				"-p", "tcp://10.168.1.2:2222|2,tcp://10.168.1.3:2222|3",
+				"-p", "tcp://10.168.1.1:2222|1,tcp://10.168.1.2:2222|2,tcp://10.168.1.3:2222|3",
```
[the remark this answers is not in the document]
agouin: Flags are switched around a bit now. 
- `--peers`/`-p` replaced by string slice `--cosigner`/`-c`. The delimiter is removed. The order in which the `-c` flags are used determines shard ID 1->n.
- `args[0]` csv for chain nodes replaced by string slice flag `--node`/`-n`. This makes it easier to string `-n` flags together for multiple chain node `privValAddr`.
`--mode` (string, either default-`threshold` or `single`) replaces `-c`.

cmd/horcrux/cmd/cosigner.go
```diff
@@ -143,36 +143,39 @@ func startCosignerCmd() *cobra.Command {
...
-			cosigners := []signer.Cosigner{}
+			cosignerConfig := config.Config.CosignerConfig
 
-			// add ourselves as a peer so localcosigner can handle GetEphSecPart requests
-			peers := []signer.CosignerPeer{{
-				ID:        key.ID,
-				PublicKey: key.RSAKey.PublicKey,
-			}}
+			remoteCosigners := make([]signer.Cosigner, 0, len(cosignerConfig.Peers)-1)
+			peers := make([]signer.CosignerPeer, len(cosignerConfig.Peers))
```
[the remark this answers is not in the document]
agouin: yes, this is validated in `ValidateThresholdModeConfig`

cmd/horcrux/cmd/testdata/config-migrated.yaml
```diff
@@ -1,8 +1,8 @@
 cosigner:
   threshold: 2
-  shares: 3
-  p2p-listen: tcp://127.0.0.1:2222
   peers:
+  - share-id: 1
+    p2p-addr: tcp://127.0.0.1:2222
```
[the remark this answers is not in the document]
agouin: from our discussion off GH, we talked about having these be `shards` instead of `peers`, with an `id` property within each. After playing around with it some more I changed it to `cosigners` instead of `peers`, and changed `share-id` -> `shardID` within. I changed the higher level `cosigner` property to `thresholdMode`, since I think it better represents the configuration that is specific to using `threshold` mode. The grouping of `shardID` and `p2pAddr` adequately represents what cosigners need to know about each other, the network reachable address and the key shard ID for the ephemeral shares (nonce) and combined signature.

signer/config.go
```diff
@@ -198,26 +194,43 @@ type CosignerPeerConfig struct {
...
 		if peer.ShareID < 1 || peer.ShareID > shares {
-			return fmt.Errorf("peer ID %v in args is out of range, must be between 1 and %v, inclusive",
-				peer.ShareID, shares)
+			errs = append(errs, fmt.Errorf("peer ID %d in args is out of range, must be between 1 and %d, inclusive",
+				peer.ShareID, shares))
+		}
+
+		url, err := url.Parse(peer.P2PAddr)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to parse peer %d p2p address: %w", peer.ShareID, err))
```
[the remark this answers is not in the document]
agouin: peer now replaced by cosigner throughout codebase. (peer ID is replaced by cosigner shard ID)

signer/config_test.go
```diff
@@ -93,10 +93,12 @@ func TestValidateCosignerConfig(t *testing.T) {
 			config: signer.Config{
 				CosignerConfig: &signer.CosignerConfig{
 					Threshold: 2,
-					Shares:    3,
 					Timeout:   "1000ms",
-					P2PListen: "tcp://127.0.0.1:2222",
 					Peers: signer.CosignerPeersConfig{
+						{
+							ShareID: 1,
```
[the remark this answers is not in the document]
agouin: the cosigner shard ID is required for the ed25519 threshold signing process, so the int 1->n is needed.

signer/config_test.go
agouin: updated to remove this requirement

signer/threshold_validator.go
agouin: shrunk

DavidNix (approved):

PR #149, opened 2023-05-15
tidy up usage and flags

agouin:
`horcrux address`
- give a more helpful error including the chain ID  if `horcrux address <chain-id>` fails to find they key

`horcrux config init`
- `[chain-nodes]` is not an optional argument, it is required with `--node` flag(s)
- clarify `--node` and `--cosigner` flags.

`horcrux create-rsa-shards`
- `shards` is not an argument, but a required flag

`horcrux create-ed25519-shards`
- all flags are required

`horcrux state set`
- `chain-id` and `height` are required args

`horcrux state import`
- `chain-id` is a required arg

### DIFF
--- a/cmd/horcrux/cmd/config.go
+++ b/cmd/horcrux/cmd/config.go
@@ -125,7 +125,8 @@ for threshold signer mode, --cosigner flags and --threshold flag are required.
 
 	f := cmd.Flags()
 	f.StringP(flagSignMode, "m", string(signer.SignModeThreshold),
-		`sign mode, "threshold" (recommended) or "single" (unsupported). threshold mode requires --cosigners and --threshold`,
+		"sign mode, \"threshold\" (recommended) or \"single\" (unsupported). "+
+			"threshold mode requires --cosigner flags and --threshold flag",
 	)
 	f.StringSliceP(flagNode, "n", []string{}, "chain nodes in format tcp://{node-addr}:{privval-port} \n"+
 		"(e.g. --node tcp://sentry-1:1234 --node tcp://sentry-2:1234 --node tcp://sentry-3:1234 )")
--- a/cmd/horcrux/cmd/config.go
+++ b/cmd/horcrux/cmd/config.go
@@ -17,6 +17,7 @@ const (
 	flagRaftTimeout = "raft-timeout"
 	flagGRPCTimeout = "grpc-timeout"
 	flagOverwrite   = "overwrite"
+	flagBare        = "bare"
 )
 
 func configCmd() *cobra.Command {
@@ -43,6 +44,7 @@ for threshold signer mode, --cosigner flags and --threshold flag are required.
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			cmdFlags := cmd.Flags()
 
+			bare, _ := cmdFlags.GetBool(flagBare)
 			nodes, _ := cmdFlags.GetStringSlice(flagNode)
 
 			cn, err := signer.ChainNodesFromFlag(nodes)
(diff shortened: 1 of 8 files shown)


cmd/horcrux/cmd/config.go
```diff
@@ -122,23 +122,31 @@ func initCmd() *cobra.Command {
...
+			"threshold mode requires --cosigner flags and --threshold flag",
 	)
-	cmd.Flags().StringSliceP(flagNode, "n", []string{}, "chain nodes in format tcp://{p2p-addr}:{port}")
-	cmd.Flags().StringSliceP(flagCosigner, "c", []string{},
-		"cosigners in format tcp://{p2p-addr}:{port}|{shard-id} \n"+
-			"(i.e. \"tcp://node-1:2222|1,tcp://node-2:2222|2,tcp://node-3:2222|3\")")
-	cmd.Flags().IntP(flagThreshold, "t", 0, "number of shards required for threshold signature")
-	cmd.Flags().StringP(
+	f.StringSliceP(flagNode, "n", []string{}, "chain nodes in format tcp://{node-addr}:{privval-port} \n"+
+		"(e.g. --node tcp://sentry-1:1234 --node tcp://sentry-2:1234 --node tcp://sentry-3:1234 )")
```
DavidNix, line 132: The more idiomatic way is to use backticks. 

```go
`chain nodes in format tcp://{node-addr}:{privval-port}
(e.g. --node tcp://sentry-1:1234 --node tcp://sentry-2:1234 --node tcp://sentry-3:1234 )`
```

No need to escape newlines.

DavidNix: I think you have an extra space before the close paren `)`.

cmd/horcrux/cmd/config.go
```diff
@@ -33,12 +33,12 @@ func configCmd() *cobra.Command {
 
 func initCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "init [chain-nodes]",
+		Use:     "init",
```
DavidNix, line 36: Instead of all these flags, I wonder if `init` generates an example that the user must then hand-edit? 

That way, we can simply delete all these flags and make this subcommand simpler. 

For our own deployments, we likely won't use this subcommand anymore. Rather, we'll create the config file as a configmap and mount it into the container.

agouin: Maybe a `--bare` flag to generate an empty config (or disable validation) would be helpful? Right now this command enforces that the final config is valid.

cmd/horcrux/cmd/config.go
```diff
@@ -122,23 +122,31 @@ func initCmd() *cobra.Command {
 			return nil
 		},
 	}
-	cmd.Flags().StringP(flagSignMode, "m", string(signer.SignModeThreshold),
-		`sign mode, "threshold" (recommended) or "single" (unsupported). threshold mode requires --cosigners and --threshold`,
+
+	f := cmd.Flags()
+	f.StringP(flagSignMode, "m", string(signer.SignModeThreshold),
+		"sign mode, \"threshold\" (recommended) or \"single\" (unsupported). "+
+			"threshold mode requires --cosigner flags and --threshold flag",
```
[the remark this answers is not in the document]
agouin: yes, it was the linter. I added `// nolint` for now

cmd/horcrux/cmd/config.go
DavidNix: Nah, I'd keep it as-is then instead of adding more complexity. It was just a thought, not a blocker.

DavidNix (approved):

DavidNix (approved):

DavidNix (approved):

PR #153, opened 2023-05-23
v3 polish

agouin:
- Create state dir on startup if doesn't exist. Avoids the need to do this manually
- Use dash for cosigner dirs

### DIFF
--- a/docs/migrating.md
+++ b/docs/migrating.md
@@ -95,21 +95,21 @@ Horcrux uses RSA 4096 keys to encrypt cosigner-to-cosigner p2p communication. Th
 
 ```bash
 $ horcrux create-rsa-shards --shards 3
-Created RSA Shard cosigner-1/rsa_keys.json
-Created RSA Shard cosigner-2/rsa_keys.json
-Created RSA Shard cosigner-3/rsa_keys.json
+Created RSA Shard cosigner_1/rsa_keys.json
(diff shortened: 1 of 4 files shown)


docs/migrating.md
```diff
@@ -95,21 +95,21 @@ Horcrux uses RSA 4096 keys to encrypt cosigner-to-cosigner p2p communication. Th
 
 ```bash
 $ horcrux create-rsa-shards --shards 3
-Created RSA Shard cosigner_1/rsa_keys.json
-Created RSA Shard cosigner_2/rsa_keys.json
-Created RSA Shard cosigner_3/rsa_keys.json
+Created RSA Shard cosigner-1/rsa_keys.json
+Created RSA Shard cosigner-2/rsa_keys.json
+Created RSA Shard cosigner-3/rsa_keys.json
```
DavidNix, line 100: Is there a reason `rsa_keys` use underscore instead of dash?

DavidNix (approved):
I think underscore for dir naming was fine. I'm curious why the change? 

But, it's fine either way. I can get onboard with dash instead of underscore.

agouin:
> I'm curious why the change?

@danbryan pointed out that this is not consistent with our secret naming in the horcrux deployments. But, I think you're right, the underscore is fine, and prevents changes to existing v3 rc users/deployments.

PR #157, opened 2023-06-07
Make Horcrux a static executable

### LINKED ISSUE #154: v3 breaks on Ubuntu 20.04
Not sure if this is intended, but v3 is built on latest Ubuntu release and so uses the latest GLIBC (2.35) which is not supported by 20.04 (2.31). Might not be supported by other distros as well.

```
./horcrux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./horcrux)
./horcrux: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./horcrux)
```

vimystic:
Eliminating dependencies on system libraries by making the Horcrux binary static. 
This will fix #154

### DIFF
--- a/goreleaser.yaml
+++ b/goreleaser.yaml
@@ -16,6 +16,8 @@ builds:
     goarch:
       - amd64
       - arm64
+    env:
+      - CGO_ENABLED=0  
 
 checksum:
   name_template: SHA256SUMS-{{.Version}}.txt


agouin (approved):
Good find!

PR #160, opened 2023-06-17
nonce optimization

agouin:
Generate nonces concurrently

Swap for coinbase kryptology for threshold ed25519

PR #164, opened 2023-06-28
set and get nonces concurrently

agouin:
phase 1 of #160  and #162

### DIFF
--- a/signer/local_cosigner.go
+++ b/signer/local_cosigner.go
@@ -426,29 +426,32 @@ func (cosigner *LocalCosigner) GetEphemeralSecretParts(
 	id := cosigner.GetID()
 
 	var eg errgroup.Group
+	// getting nonces requires encrypting and signing for each cosigner,
+	// so we perform these operations in parallel.
 
-	for i, pubKey := range cosigner.rsaPubKeys {
-		if pubKey.ID == id {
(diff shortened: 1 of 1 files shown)


signer/local_cosigner.go
```diff
@@ -419,30 +420,45 @@ func (cosigner *LocalCosigner) GetEphemeralSecretParts(
...
+				Round:     hrst.Round,
+				Step:      hrst.Step,
+				Timestamp: time.Unix(0, hrst.Timestamp),
+			})
+
+			if i > id {
+				res.EncryptedSecrets[i-2] = secretPart
+			} else {
+				res.EncryptedSecrets[i-1] = secretPart
+			}
```
mark-rushakoff, line 452: This threw me off for a moment since it wasn't immediately clear that `i` was map keys -- it seemed that there was an unhandled `i == 0`  case that would lead to indexing a slice at -1.

It could maybe be clearer with `idx := i-1` instead of `i := i`, but this is narrow enough in scope that I don't think it really needs to change.

mark-rushakoff (approved):
Wouldn't hurt to leave some comments mentioning why it's worth spreading this work across CPUs, but this seems fine as far as I can see.

signer/local_cosigner.go
agouin: good call, updated

PR #165, opened 2023-06-28
nonce rename

agouin:
- rename EphemeralSecretPart -> Nonce
- rename EphemeralPublic -> NoncePublic
- other minor renames
- improve some comments


signer/metrics.go
```diff
@@ -134,7 +134,7 @@ var (
 		Help: "Seconds Since Last Local Finish Sign (Should stay below 2 * Block Time)",
 	})
 
-	secondsSinceLastLocalEphemeralShareTime = promauto.NewGauge(prometheus.GaugeOpts{
+	secondsSinceLastLocalNonceTime = promauto.NewGauge(prometheus.GaugeOpts{
 		Name: "signer_seconds_since_last_local_ephemeral_share_time",
```
mark-rushakoff, line 138: Is it too late to change the prometheus metric names? Do they need to be changed to use `nonce` too?

mark-rushakoff (approved):
Looks like a straightforward rename.

signer/metrics.go
agouin: I avoided these since it would be a breaking change. We aren't yet consuming these in our metrics stack but other validators are.

PR #169, opened 2023-07-07
sign state signaling

agouin:
Use sync.Cond for sign state signaling

If a block sign process has been initiated, other sign requests for the same height will wait until that process has completed, and need to be notified that they can proceed to check the caches for the existing signature.

Previously, this was done with a 10ms wait loop. `sync.Cond` gives us a way to notify all of the waiting goroutines.

### DIFF
--- a/signer/sign_state.go
+++ b/signer/sign_state.go
@@ -51,17 +51,19 @@ func ProposalToStep(_ *cometproto.Proposal) int8 {
 
 // SignState stores signing information for high level watermark management.
 type SignState struct {
-	Height      int64                         `json:"height"`
-	Round       int64                         `json:"round"`
-	Step        int8                          `json:"step"`
-	NoncePublic []byte                        `json:"nonce_public"`
-	Signature   []byte                        `json:"signature,omitempty"`
-	SignBytes   cometbytes.HexBytes           `json:"signbytes,omitempty"`
-	cache       map[HRSKey]SignStateConsensus `json:"-"`
-	mu          sync.RWMutex                  `json:"-"`
-	cond        *sync.Cond                    `json:"-"`
+	Height      int64               `json:"height"`
+	Round       int64               `json:"round"`
+	Step        int8                `json:"step"`
+	NoncePublic []byte              `json:"nonce_public"`
+	Signature   []byte              `json:"signature,omitempty"`
+	SignBytes   cometbytes.HexBytes `json:"signbytes,omitempty"`
 
-	filePath string `json:"-"`
+	filePath string
+
+	// mu protects the cache and is used for signaling with cond.
+	mu    sync.RWMutex
+	cache map[HRSKey]SignStateConsensus
+	cond  *sync.Cond
 }
 
 func (signState *SignState) existingSignatureOrErrorIfRegression(hrst HRSTKey, signBytes []byte) ([]byte, error) {
--- a/signer/sign_state.go
+++ b/signer/sign_state.go
@@ -21,7 +21,6 @@ const (
 	stepPrevote   int8 = 2
 	stepPrecommit int8 = 3
 	blocksToCache      = 3
-	dontSave           = "DONT_SAVE"
 )
 
 func CanonicalVoteToStep(vote *cometproto.CanonicalVote) int8 {
@@ -229,7 +228,7 @@ func (signState *SignState) Save(
 // Save persists the FilePvLastSignState to its filePath.
 func (signState *SignState) save(jsonBytes []byte) {
 	outFile := signState.filePath
-	if outFile == dontSave {
+	if outFile == os.DevNull {
 		return
 	}
 	if outFile == "" {
--- a/signer/sign_state.go
+++ b/signer/sign_state.go
@@ -51,15 +51,53 @@ func ProposalToStep(_ *cometproto.Proposal) int8 {
 
 // SignState stores signing information for high level watermark management.
 type SignState struct {
-	Height      int64               `json:"height"`
-	Round       int64               `json:"round"`
-	Step        int8                `json:"step"`
-	NoncePublic []byte              `json:"ephemeral_public"`
-	Signature   []byte              `json:"signature,omitempty"`
-	SignBytes   cometbytes.HexBytes `json:"signbytes,omitempty"`
-	cache       map[HRSKey]SignStateConsensus
+	Height      int64                         `json:"height"`
+	Round       int64                         `json:"round"`
+	Step        int8                          `json:"step"`
+	NoncePublic []byte                        `json:"nonce_public"`
+	Signature   []byte                        `json:"signature,omitempty"`
+	SignBytes   cometbytes.HexBytes           `json:"signbytes,omitempty"`
+	cache       map[HRSKey]SignStateConsensus `json:"-"`
+	mu          sync.RWMutex                  `json:"-"`
+	cond        *sync.Cond                    `json:"-"`
 
-	filePath string
+	filePath string `json:"-"`
+}
+
+func (signState *SignState) existingSignatureOrErrorIfRegression(hrst HRSTKey, signBytes []byte) ([]byte, error) {
+	signState.mu.RLock()
+	defer signState.mu.RUnlock()
+
+	sameHRS, err := signState.CheckHRS(hrst)
+	if err != nil {
+		return nil, err
+	}
+
+	if !sameHRS {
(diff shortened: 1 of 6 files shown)


mark-rushakoff (commented):
I've left a few comments, but I'm going to have to return to this next week to finish reviewing.

signer/sign_state.go
```diff
@@ -51,15 +51,53 @@ func ProposalToStep(_ *cometproto.Proposal) int8 {
...
-
-	filePath string
+	Height      int64                         `json:"height"`
+	Round       int64                         `json:"round"`
+	Step        int8                          `json:"step"`
+	NoncePublic []byte                        `json:"nonce_public"`
+	Signature   []byte                        `json:"signature,omitempty"`
+	SignBytes   cometbytes.HexBytes           `json:"signbytes,omitempty"`
+	cache       map[HRSKey]SignStateConsensus `json:"-"`
+	mu          sync.RWMutex                  `json:"-"`
```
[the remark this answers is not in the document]
agouin: good call, updated and added a comment.

signer/sign_state.go
```diff
@@ -138,33 +182,40 @@ func (signState *SignState) Save(
...
 	signState.SignBytes = ssc.SignBytes
+
+	jsonBytes, err := cometjson.MarshalIndent(signState, "", "  ")
+	if err != nil {
+		panic(err)
+	}
+
+	signState.mu.Unlock()
+
+	signState.cond.Broadcast()
```
mark-rushakoff, line 193: Since `sync.Cond` is a rarely used part of the standard library, can you add a comment here explaining why/what we are broadcasting here?

signer/threshold_validator.go
```diff
@@ -103,12 +103,90 @@ func (pv *ThresholdValidator) mustLoadChainState(chainID string) ChainSignState
...
+	err := css.lastSignStateInitiated.Save(NewSignStateConsensus(height, round, step), &pv.pendingDiskWG)
+	if err == nil {
+		// good to sign
+		return nil, time.Time{}, nil
+	}
+
+	existingSignature, existingTimestamp, sameBlockErr := pv.getExistingBlockSignature(chainID, block)
+
+	if _, ok := err.(*SameHRSError); ok {
+		// Wait for last sign state signature to be the same block
```
mark-rushakoff, line 120: I don't think this comment fits here; I think it's the overall logic of this block, but it looks like it details the double-error check adjacent to the comment.

Separate with a blank line, or add a new comment about why the errors are checked in this particular way?

signer/threshold_validator.go
```diff
@@ -103,12 +103,90 @@ func (pv *ThresholdValidator) mustLoadChainState(chainID string) ChainSignState
...
+		if sameBlockErr == nil {
+			if existingSignature != nil {
+				return existingSignature, existingTimestamp, nil
+			}
+			// good to sign again
+			return nil, time.Time{}, nil
+		}
+
+		if _, ok := sameBlockErr.(*StillWaitingForBlockError); !ok {
+			return nil, existingTimestamp, sameBlockErr
```
mark-rushakoff, line 130: Would it be better to wrap the error with more detail here (`fmt.Errorf("...: %w", sameBlockErr)`)?

signer/threshold_validator.go
```diff
@@ -103,12 +103,90 @@ func (pv *ThresholdValidator) mustLoadChainState(chainID string) ChainSignState
...
+
+		css.lastSignState.cond.L.Lock()
+		defer css.lastSignState.cond.L.Unlock()
+		for {
+			css.lastSignState.cond.Wait()
+
+			ssc, ok := css.lastSignState.cache[block.HRSKey()]
+			if !ok {
+				pv.logger.Debug(
+					"Cond does not exist in cache",
```
mark-rushakoff, line 144: Is it actually `Cond` that doesn't exist in the cache here?

Maybe the log message should note that it is waiting for the next sign state signal, for a bit more detail for the human log reader.

signer/threshold_validator.go
```diff
@@ -103,12 +103,90 @@ func (pv *ThresholdValidator) mustLoadChainState(chainID string) ChainSignState
...
+		}
+	} else {
+		if sameBlockErr == nil {
+			return existingSignature, block.Timestamp, nil
+		}
+		return nil, existingTimestamp, pv.newBeyondBlockError(chainID, block.HRSKey())
+	}
+}
+
+// HandleError will alert any waiting goroutines that an error
```
mark-rushakoff, line 176: s/HandleError/notifyBlockSignError/

signer/threshold_validator.go
```diff
@@ -369,20 +464,34 @@ func (pv *ThresholdValidator) LoadSignStateIfNecessary(chainID string) error {
...
-	latestBlock, existingSignature := css.lastSignState.GetFromCache(
-		hrs,
-		css.lastSignStateMutex,
-	)
+	latestBlock, existingSignature := css.lastSignState.GetFromCache(block.HRSKey())
+
+	return pv.compareBlockSignature(chainID, block, latestBlock, existingSignature)
+}
+
+func (pv *ThresholdValidator) compareBlockSignature(
```
mark-rushakoff, line 474: Can you add godoc for this so other readers understand what the return values mean?

signer/threshold_validator.go
```diff
@@ -457,44 +567,13 @@ func (pv *ThresholdValidator) SignBlock(chainID string, block *Block) ([]byte, t
...
-				Step:   step,
-			}
-			return nil, existingTimestamp, pv.newBeyondBlockError(chainID, hrs)
-		}
+	existingSignature, existingTimestamp, err := pv.SaveLastSignedStateInitiated(chainID, block)
+	if err != nil {
+		return nil, stamp, err
+	}
+	if existingSignature != nil {
+		pv.logger.Debug("Returning existing signature", "signature", fmt.Sprintf("%x", existingSignature))
```
mark-rushakoff, line 575: Note here, the `fmt.Sprintf` call will be eagerly evaluated, regardless of whether the logger is debug-level enabled.

If this is an infrequently executed code path -- and I think that's the case here -- that is fine.

mark-rushakoff (commented):
I left a bunch more comments here.

I've spent probably over two hours reading this at this point, and I am still finding the changes difficult to follow; so I suggested several places where clarifying comments would help the reader quite a bit. 

If I were more familiar with the code base here, I would probably be further along comprehending these changes in the amount of time I've read them. Take it for what it's worth.

signer/sign_state.go
```diff
@@ -138,33 +182,40 @@ func (signState *SignState) Save(
...
 	}
 
 	return nil
 }
 
 // Save persists the FilePvLastSignState to its filePath.
-func (signState *SignState) save() {
+func (signState *SignState) save(jsonBytes []byte) {
 	outFile := signState.filePath
 	if outFile == "none" {
```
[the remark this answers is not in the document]
agouin: yes, /dev/null works well to remove this special handling

signer/threshold_validator.go
```diff
@@ -103,12 +103,90 @@ func (pv *ThresholdValidator) mustLoadChainState(chainID string) ChainSignState
...
+			}
+			if _, ok := sameBlockErr.(*StillWaitingForBlockError); !ok {
+				return nil, existingTimestamp, sameBlockErr
+			}
+
+			pv.logger.Debug(
+				"Stil waiting for block to be signed",
+				"height", height,
+				"round", round,
+				"step", step,
```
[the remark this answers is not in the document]
agouin: Added information about the latest signed block to the log

signer/sign_state.go
```diff
@@ -263,9 +294,12 @@ func (signState *SignState) FreshCache() *SignState {
 		Signature:   signState.Signature,
 		SignBytes:   signState.SignBytes,
 		cache:       make(map[HRSKey]SignStateConsensus),
-		filePath:    signState.filePath,
+
+		filePath: signState.filePath,
 	}
 
+	newSignState.cond = sync.NewCond(&newSignState.mu)
```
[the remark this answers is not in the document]
agouin: no, `FreshCache` is used to create a clone of a  `SignState` with a fresh cache and new mutex and cond. It is used only during instantiation of the `lastSignState` and `lastSignStateInitiated`.

mark-rushakoff (approved):
The cond comments helped me out in reading this. I think there are still some existing subtleties here, but I don't have any more suggestions on how to reduce them at the moment.

PR #174, opened 2023-07-15
Interchaintest for e2e tests

agouin:
ports e2e framework to use interchaintest.

use test matrix in CI, removing need for self hosted test runner.

improves test coverage by using slashing period of 10 blocks, 90% uptime required, and reducing number of validators in test from 4 to 2, which requires the horcrux validator for consensus.

image


go.work
```diff
@@ -0,0 +1,6 @@
+go 1.20
+
+use (
+	.
+	./test
+)
```
DavidNix, line 6: I think the Go workspace is to support the separate go.mod under the `test` directory. Why not just use 1 go.mod for the whole project?

test/validator.go
```diff
@@ -1,197 +1,273 @@
...
+
+	signerImage        = "horcrux-test"
+	binary             = "horcrux"
+	signerImageUidGid  = "2345:2345"
+	signerImageHomeDir = "/home/horcrux"
+)
+
+// chainWrapper holds the initial configuration for a chain to start from genesis.
+type chainWrapper struct {
+	chain           **cosmos.CosmosChain
```
DavidNix, line 44: The double star (pointer to a pointer) is uncommon. Have you tried using the `chainWrapper` strust as a pointer when callers use it? This may eliminate the double pointer.

test/validator.go
```diff
@@ -1,197 +1,273 @@
...
 
-// Name is the name of the test validator
-func (tv *Validator) Name() string {
-	return fmt.Sprintf("validator-%d-%s", tv.Index, tv.tl.Name())
-}
+	err = ic.Build(ctx, nil, interchaintest.InterchainBuildOptions{
+		TestName:  t.Name(),
+		Client:    client,
+		NetworkID: network,
+		// BlockDatabaseFile: interchaintest.DefaultBlockDatabaseFilepath(),
```
DavidNix, line 94: nit: Can you add a comment informing a reader why we're commenting this out?

test/validator.go
```diff
@@ -1,197 +1,273 @@
...
-		Threshold: threshold,
-		PubKeys:   make(map[string]cometcrypto.PubKey),
+// startChains starts the given chains locally within docker composed of containers.
+func startChains(
+	ctx context.Context,
+	t *testing.T,
+	logger *zap.Logger,
+	client *client.Client,
+	network string,
+	chains ...chainWrapper,
```
DavidNix, line 58: Changing this to `chains ...*chainWrapper` may help clean up some of the dereferncing.

test/validator_threshold.go
```diff
@@ -0,0 +1,338 @@
...
+func preGenesisAllHorcruxThreshold(
+	ctx context.Context,
+	logger *zap.Logger,
+	client *client.Client,
+	network string,
+	totalSigners int, // total number of signers for the single horcrux validator
+	threshold uint8, // key shard threshold, and therefore how many horcrux signers must participate to sign a block
+	sentriesPerValidator int, // how many sentries for each horcrux validator (min: sentriesPerSigner, max: totalSentries)
+	sentriesPerSigner int, // how many sentries should each horcrux signer connect to (min: 1, max: sentriesPerValidator)
+	chain **cosmos.CosmosChain,
```
DavidNix, line 117: I think we can simplify to a single star/pointer.

DavidNix (commented):
It's great we're getting more test coverage on CI!

agouin:
> It's great we're getting more test coverage on CI!

All of the e2e tests were running as part of CI, but on a single job on the external runner, which made figuring out which specific test was failing more difficult.

go.work
agouin: To avoid the interchaintest and it's dependencies as dependencies for the horcrux binary. This minimizes required deps to download for make install for docker and release, for example.

test/validator.go
agouin: The double star removed the need for double wrapped functions for pre-genesis, but I updated it so that it's the single pointer.

test/validator.go
agouin: we won't ever need this in horcrux I don't think, so I removed it.

go.work
DavidNix: This is not how Go modules work. The compiler builds a dependency graph and is smart enough to only download dependencies used for the binary. Because the `test` package is (rightfully) never used in horcrux's `main()`; `test`'s dependencies are ignored.

I deleted the go.work and test/go.mod andI tested on [a separate branch](https://github.com/strangelove-ventures/horcrux/pull/176).

As you can see, no interchaintest modules were downloaded. 

```
❯ make install
go: downloading github.com/cosmos/cosmos-sdk v0.47.3
go: downloading google.golang.org/grpc v1.55.0
go: downloading golang.org/x/crypto v0.8.0
go: downloading github.com/cosmos/gogoproto v1.4.10
go: downloading github.com/Jille/raft-grpc-transport v1.4.0
go: downloading github.com/hashicorp/raft v1.5.0
go: downloading github.com/ecies/go/v2 v2.0.6
go: downloading golang.org/x/net v0.9.0
go: downloading google.golang.org/genproto v0.0.0-20230306155012-7f2fa6fef1f4
go: downloading go.etcd.io/bbolt v1.3.7
go: downloading github.com/decred/dcrd/dcrec/secp256k1/v4 v4.2.0
go: downloading github.com/ethereum/go-ethereum v1.11.5
go: downloading github.com/hashicorp/go-hclog v1.5.0
go: downloading golang.org/x/exp v0.0.0-20230515195305-f3d0a9c9a5cc
go: downloading github.com/fatih/color v1.13.0
go: downloading cosmossdk.io/math v1.0.1
go: downloading github.com/hdevalence/ed25519consensus v0.1.0
go: downloading github.com/cosmos/cosmos-proto v1.0.0-beta.2
go: downloading filippo.io/edwards25519 v1.0.0
```

However, I'm fine with this decision. I think documentation pointing out this decision for a go.work is useful for people unfamiliar with the codebase, though.

On my branch, I ran into a weird `go mod tidy` error and punted on it further:
```
❯ go mod tidy
go: github.com/gogo/protobuf@v1.3.3: reading github.com/gogo/protobuf/go.mod at revision v1.3.3: unknown revision v1.3.3
```

But, in theory, only one go.mod should be fine and shouldn't impact performance or maintenance.

The one caveat is if horcrux `main()` and `test` share transitive deps. Then one go.mod will use [MVS](https://go.dev/ref/mod#minimal-version-selection) which may have unintended consequences, which may be the problem with `gogo/protobuf` package above.

DavidNix (approved):

signer/remote_signer.go
```diff
@@ -62,7 +63,7 @@ func NewReconnRemoteSigner(
 
 // OnStart implements cmn.Service.
 func (rs *ReconnRemoteSigner) OnStart() error {
-	go rs.loop()
+	go rs.loop(context.Background())
```
DavidNix, line 66: Why not have `OnStart()` take a context? Seems like that context would be close to one created in `main()`.

agouin: `OnStart` is an interface method for the comet service. Removing the dependency on comet is something I'd like to do at some point, or at least removing it for the services and logging.

signer/remote_signer.go
```diff
@@ -103,7 +116,7 @@ func (rs *ReconnRemoteSigner) loop() {
 				conn = nil
 				rs.Logger.Error("Secret Conn", "err", err)
 				rs.Logger.Info("Retrying", "sleep (s)", 3, "address", rs.address)
-				time.Sleep(time.Second * 3)
+				time.Sleep(time.Second * 1)
```
DavidNix, line 119: A ticker may be more idiomatic. It could remove the need to call `Sleep` twice.

Fyi, `time.Sleep` is rare for idiomatic Go - usually indicates a code smell or hack.

go.work
agouin: The protobuf is due to the replace statement in interchaintest. We would need to copy all of the replaces also which are currently only in the test/go.mod.

Good to know about the dependency graph. I thought we would have a performance impact [here](https://github.com/strangelove-ventures/horcrux/blob/andrew/interchaintest/docker/horcrux/native.Dockerfile#L7-L9) since it prefetches go deps for docker caching, but I could be wrong.

signer/remote_signer.go
agouin: Yes, been meaning to clean this one up too. It's an old copy paste from the tendermint code base.

go.work
DavidNix: Oh yes, it might fetch all in that case.

DavidNix: But that step may not be necessary since `go build` downloads deps anyway. But we may be trying to optimize docker builds in some fashion.

signer/remote_signer.go
agouin: done

PR #178, opened 2023-07-18
use go-ethereum ecies due to instability with ecies/go/v2

agouin:
The new interchaintest e2e suite caught a bug with the `ecies/go/v2` lib. It can fail to encrypt properly when used concurrently. 

Swaps out ecies functionality with go ethereum lib instead, and tests updated to catch the bug in the future.


cmd/horcrux/cmd/leader_election.go
```diff
@@ -107,26 +107,40 @@ func getLeaderCmd() *cobra.Command {
...
-			if err != nil {
-				return err
-			}
+			var id int
 
-			key, err := signer.LoadCosignerRSAKey(keyFile)
-			if err != nil {
-				return fmt.Errorf("error reading cosigner key (%s): %w", keyFile, err)
+			keyFile, err := config.KeyFileExistsCosignerECIES()
+			if err == nil {
```
DavidNix, line 113: We are swallowing the error on line 112. Do we want to log it somewhere? But I see you're trying to determine if there's an ECIES or RSA key, preferring ECIES first.

Checking for `err == nil` is unusual imo. `err != nil` is what Go devs typically expect, although this may be cause for an exception.

signer/cosigner_security_ecies_test.go
```diff
@@ -58,3 +75,58 @@ func testCosignerSecurity(t *testing.T, securities []CosignerSecurity) error {
...
+						if err != nil {
+							return err
+						}
+						return nil
+					})
+				}
+				return nestedEg.Wait()
+			})
+		}
+		require.NoErrorf(b, eg.Wait(), "success count: %d", i)
```
DavidNix, line 130: It's unusual to see an errgroup (especially 2) within a benchmark test. I'm sure @mark-rushakoff may have some thoughts here. Why not run everything serially? 

Additionally, splitting benchmark tests between the methods `EncryptAndSign` and `DecryptAndVerify` will help isolate which methods become slower or faster over time.

Subtests may be useful here. https://go.dev/blog/subtests (Or may not 😄 )

signer/threshold_validator.go
```diff
@@ -347,27 +348,26 @@ func (pv *ThresholdValidator) waitForPeerNonces(
...
 
-	// Check so that getEphemeralWaitGroup.Done is not called more than (threshold - 1) times which causes hardlock
+	// Check so that wg.Done is not called more than (threshold - 1) times which causes hardlock
 	thresholdPeersMutex.Lock()
-	if len(encryptedNoncesThresholdMap) < pv.threshold-1 {
-		(encryptedNoncesThresholdMap)[peer] = ephemeralSecretParts.Nonces
+	if len(nonces) < pv.threshold-1 {
+		nonces[peer] = peerNonces.Nonces
 		defer wg.Done()
 	}
```
DavidNix, line 372: This is old code but `defer wg.Done()` conditionally (after an `if`) makes me thing there's a chance for a deadlock. Hopefully something else triggers `wg.Done()`.

signer/cosigner_security_rsa_test.go
```diff
@@ -1,41 +1,112 @@
...
+						if err != nil {
+							return err
+						}
+						return nil
+					})
+				}
+				return nestedEg.Wait()
+			})
+		}
+		require.NoErrorf(b, eg.Wait(), "success count: %d", i)
```
DavidNix, line 110: Same comment about errgroup in a benchmark test applies here.

DavidNix (approved):
Do we need to support ECIES and RSA? Can we only support ECIES? (Maybe not if we've committed to v3.) Would make things simpler.

signer/threshold_validator.go
agouin: this is done intentionally, It does `wg.Add(pv.threshold-1)` because we want to stop waiting once we have threshold responses (local cosigner included for -1). The defer is used to make sure the wg.Done happens after unlocking the mutex.

signer/cosigner_security_ecies_test.go
agouin: The errgroups were used here to simulate horcrux' runtime usage of the `CosignerSecurity`. It helped to reproduce the issue in the ecies/go/v2 lib. A benchmark may not be the best tool for this, but many thousands of iterations were required in some cases to reproduce the issue.

agouin:
> Do we need to support ECIES and RSA? Can we only support ECIES?
For backwards compatibility, and to avoid the need to rush to v4, I'd like to keep rsa support. Removing it in a future release will be nice to clean things up.

cmd/horcrux/cmd/leader_election.go
agouin: good call, cleaned this up.

signer/threshold_validator.go
mark-rushakoff: Alternate approach using channels instead of a mutex:

```go
var wg sync.WaitGroup
wgCh := make(chan *sync.WaitGroup, threshold)
for i := 0; i < threshold; i++ {
  // Seed only the threshold count worth of wait groups,
  // so we can wait on the threshold being reached, not the total.
  wgCh <- &wg
}

wg.Add(threshold)
for i := 0; i < total; i++ {
  go foo(wgCh)
}

// func foo(wgCh chan *sync.WaitGroup):
defer func() {
  if wg, ok := <-wgCh; ok {
    // Haven't called wg.Done on the threshold count yet.
    wg.Done()
  }
}
```

mark-rushakoff (approved):
I left a few small suggestions but nothing blocking merge.

go.work.sum
```diff
@@ -28,6 +28,7 @@ cloud.google.com/go/cloudtasks v1.10.0/go.mod h1:NDSoTLkZ3+vExFEWu2UJV1arUyzVDAi
 cloud.google.com/go/compute v1.14.0/go.mod h1:YfLtxrj9sU4Yxv+sXzZkyPjEyPBZfXHUvjxega5vAdo=
 cloud.google.com/go/compute v1.18.0/go.mod h1:1X7yHxec2Ga+Ss6jPyjxRxpu2uu7PLgsOVXvgU0yacs=
 cloud.google.com/go/compute v1.19.0/go.mod h1:rikpw2y+UMidAe9tISo04EHNOIf42RLYF/q8Bs93scU=
+cloud.google.com/go/compute/metadata v0.2.0/go.mod h1:zFmK7XCadkQkj6TtorcaGlCW1hT1fIilQDwofLpJ20k=
```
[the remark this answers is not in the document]
agouin: VSCode requires the go.work in order to have the IDE smarts within multiple modules without opening multiple workspaces with the go.mod at the root.

I suppose I am fine to remove and gitignore it, but it is nice for the IDE to work with everything off a fresh clone.

signer/threshold_validator.go
```diff
@@ -426,6 +430,7 @@ func (pv *ThresholdValidator) waitForPeerSetNoncesAndSign(
 		"height", hrst.Height,
 		"round", hrst.Round,
 		"step", hrst.Step,
+		"signature", hex.EncodeToString(sigRes.Signature),
```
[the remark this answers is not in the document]
agouin: Thanks for calling this out. We don't really need this log, it was more helpful for debugging. Removed.

signer/cosigner_security_ecies_test.go
agouin: Great idea, I updated them to be tests instead of benchmarks with a reasonable number of iterations.

PR #181, opened 2023-07-19
Update migrating.md

Pa1amar:


agouin:
Hi @Pa1amar ! We are cutting a release this week which replaces rsa signing with ecies, as it is much faster. RSA will still be available, but it will be deprecated and not recommended. See the deployment guide [here](https://github.com/strangelove-ventures/horcrux/blob/v3.0.0/docs/migrating.md) for the v3.0.0 release instructions.

PR #183, opened 2023-07-20
Vivek/156 : Downgrade golang version from 1.20 to 1.19

### LINKED ISSUE #156: Investigate / mitigate Increased latency on v3.0.0
After upgrading to v3.0.0, we see increased latency on the metrics `signer_cosigner_sign_lag_seconds` and `signer_cosigner_ephemeral_share_lag_seconds`. 

The latency increase is around 60% for `signer_cosigner_ephemeral_share_lag_seconds` and 30% for `signer_cosigner_sign_lag_seconds`. In the case of `signer_cosigner_sign_lag_seconds` the latency increase is the same from the local host, i.e. the leader.

We also gather regular ICMP ping latencies between our nodes, and those are stable - no change.

This is on a testnet cluster where cosigners and sentries are all colocated.

The spike/drop in regular latency (right hand column) can be disregard - related to adjusting collector to changes introduced.
 
 
Chain A:
image

Chain B (the gap filters out a different leader):
image

The setup is exactly the same - same leaders, no proxying.

vimystic:
Fixes #156 
Downgrade golang version from 1.20 to 1.19

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ private_share_*
 rsa_keys.json
 priv_validator_key.json
 dist/
+go.work.sum
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -97,7 +97,8 @@ linters-settings:
         alias: raftgrpctransport
       - pkg: github.com/hashicorp/raft-boltdb/v2
         alias: boltdb
-
+      - pkg: math/rand
+        alias: mrand
   maligned:
     suggest-new: true
   govet:
--- a/cmd/horcrux/cmd/threshold.go
+++ b/cmd/horcrux/cmd/threshold.go
@@ -11,6 +11,8 @@ import (
 	"github.com/strangelove-ventures/horcrux/signer"
 )
 
+const maxWaitForSameBlockAttempts = 3
+
 func NewThresholdValidator(
 	logger cometlog.Logger,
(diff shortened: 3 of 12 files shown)


test/validator.go
```diff
@@ -248,7 +251,7 @@ func pollForLeader(ctx context.Context, t *testing.T, cosigner *cosmos.SidecarPr
 
 // getLeader returns the current raft leader.
 func getLeader(ctx context.Context, cosigner *cosmos.SidecarProcess) (string, error) {
-	ports, err := cosigner.GetHostPorts(ctx, signerPortDocker)
+	ports, err := cosigner.GetHostPorts(ctx, signerPortDocker, debugPortDocker)
```
agouin, line 254: this change shouldn't be necessary

test/validator_threshold.go
```diff
@@ -356,3 +365,53 @@ func getSentriesForCosignerConnection(sentries cosmos.ChainNodes, numSigners int
 	}
 	return peers
 }
+func getCosingerMetrics(ctx context.Context, cosigners cosmos.SidecarProcesses) {
```
agouin, line 368: /s/getCosingerMetrics/getCosignerMetrics/

agouin (approved):

PR #184, opened 2023-07-24
waiting for same block signature - timeout and max attempts

agouin:
`sync.Cond` `Wait()` does not allow a timeout to be specified. It will wait forever until a goroutine `Broadcast()`s to wake it up.

https://gist.github.com/zviadm/c234426882bfc8acba88f3503edaaa36#file-cond2-go

This cond implementation uses a channel that can be closed, and a new channel to replace it each time, allowing a timeout to be used.


signer/threshold_validator.go
```diff
@@ -188,6 +192,11 @@ func (pv *ThresholdValidator) SaveLastSignedStateInitiated(chainID string, block
 			"latest_step", latest.Step,
 		)
 	}
+
+	return nil, existingTimestamp, fmt.Errorf(
```
mark-rushakoff, line 196: It doesn't look like there is a test covering this error case, and it seems important to ensure we properly handle the timeout case.

mark-rushakoff (approved):
Looks like there is a test case missing, but otherwise the changes look correct to me.

PR #186, opened 2023-07-27
no concurrent dial to same node

agouin:
concurrent dials can cause infinite retries.

simplifies connect loop to use context timeout only, ensuring synchronous completion before retrying next connection attempt

### DIFF
--- a/signer/remote_signer.go
+++ b/signer/remote_signer.go
@@ -102,6 +102,7 @@ func (rs *ReconnRemoteSigner) loop(ctx context.Context) {
 			return
 		}
 
+		retries := 0
 		for conn == nil {
 			var err error
 			conn, err = rs.establishConnection(ctx)
@@ -111,12 +112,14 @@ func (rs *ReconnRemoteSigner) loop(ctx context.Context) {
 				break
 			}
 
-			sentryConnectTries.Add(float64(1))
+			sentryConnectTries.Add(1)
 			totalSentryConnectTries.Inc()
+			retries++
 			rs.Logger.Error(
 				"Error establishing connection, will retry",
 				"sleep (s)", connRetrySec,
 				"address", rs.address,
+				"attempt", retries,
 				"err", err,
 			)
 		}
--- a/signer/remote_signer.go
+++ b/signer/remote_signer.go
@@ -105,9 +105,11 @@ func (rs *ReconnRemoteSigner) loop(ctx context.Context) {
 		retries := 0
 		for conn == nil {
 			var err error
+			timer := time.NewTimer(connRetrySec * time.Second)
 			conn, err = rs.establishConnection(ctx)
 			if err == nil {
 				sentryConnectTries.Set(0)
+				timer.Stop()
 				rs.Logger.Info("Connected to Sentry", "address", rs.address)
 				break
 			}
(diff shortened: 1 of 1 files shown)


signer/remote_signer.go
```diff
@@ -73,98 +75,82 @@ func (rs *ReconnRemoteSigner) OnStop() {
...
+		for conn == nil {
+			var err error
+			conn, err = rs.establishConnection(ctx)
+			if err == nil {
+				sentryConnectTries.Set(0)
+				rs.Logger.Info("Connected to Sentry", "address", rs.address)
+				break
 			}
+
+			sentryConnectTries.Add(float64(1))
```
mark-rushakoff, line 114: Correct but `.Add(1)` is also correct and shorter too.

signer/remote_signer.go
```diff
@@ -73,98 +75,82 @@ func (rs *ReconnRemoteSigner) OnStop() {
...
-			go rs.attemptConnection(ctx, &cw, connected)
-			select {
-			case <-connected:
-				break ConnLoop
-			case <-ticker.C:
-				cancel()
-				rs.Logger.Info("Retrying", "sleep (s)", 1, "address", rs.address)
+		for conn == nil {
+			var err error
+			conn, err = rs.establishConnection(ctx)
```
mark-rushakoff, line 107: I'm not sure if a backoff would help here. For a timeout-driven failure mode, I would usually expect to see an increasing timeout. On the other hand, two seconds is already gratuitous; if it fails in 2 seconds I don't think increasing it will help.

signer/remote_signer.go
```diff
@@ -73,98 +75,82 @@ func (rs *ReconnRemoteSigner) OnStop() {
...
+			if err == nil {
+				sentryConnectTries.Set(0)
+				rs.Logger.Info("Connected to Sentry", "address", rs.address)
+				break
 			}
+
+			sentryConnectTries.Add(float64(1))
+			totalSentryConnectTries.Inc()
+			rs.Logger.Error(
```
mark-rushakoff, line 116: In case this is happening repeatedly, a counter (associated with our iteration in the `for conn == nil` loop) might help in reading the logs. But maybe it doesn't make a difference if it's happened 10 times or 1,000 times.

mark-rushakoff (approved):

signer/remote_signer.go
```diff
@@ -18,6 +18,8 @@ import (
 	cometproto "github.com/cometbft/cometbft/proto/tendermint/types"
 )
 
+const connRetrySec = 2
```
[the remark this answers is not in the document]
DavidNix: +1

DavidNix (approved):

signer/remote_signer.go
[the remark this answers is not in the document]
agouin: it's also logging out the seconds as an int which was why I separated it.

signer/remote_signer.go
agouin: tendermint has a 3 second window for connections when the privval interface is enabled. If a remote signer does not connect during that timeframe, it panics, so a backoff beyond 2 seconds wouldn't work well unfortunately.

signer/remote_signer.go
agouin: thanks, was a copy paste.

signer/remote_signer.go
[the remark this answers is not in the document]
DavidNix: `%s` and `%v` will log it out in a human friendly format. 

```go
	const dur = 2 * time.Second
	fmt.Printf("%s\n", dur)
	fmt.Printf("%v\n", dur)
```

Both log "2s".

PR #187, opened 2023-08-01
Remove goreleaser

agouin:
Copy static binaries from docker images instead since go-ethereum/crypto/secp256k1 requires cgo


vimystic (approved):
lgtm : pending tests completion.

PR #194, opened 2023-09-02
Update migrating.md

rub-vs:
TAG version changed to latest binary
wget changed address to new download address for binary no need to extract file since binary is downloaded directly no need to delete files that would have been in packages since they are no longer there


agouin (approved):

agouin:
Thank you!

PR #198, opened 2023-09-11
FEAT: Add per chain and node metrics

Gustavobelfort:
## Summary
This Pull Request introduces per-chain and per-node metrics collection into the Horcrux monitoring system. The main objective of this update is to enhance system observability, enabling a deeper insight into the individual performance of each node and chain. This is instrumental for network health assessment, debugging, and performance optimization.

## Background
Prior to this PR, Horcrux only provided global metrics which made it challenging to debug issues on a per-node or per-chain basis. As Horcrux networks can be composed of multiple chains and nodes, a granular level of metrics is essential for operational excellence.

## Changes
- New Labels collection in the metrics that might pertain to multiple chains and/or signer (sentry) nodes

---

As this is an extension of the existing metrics collection, it does not require special migration steps. However, due to the label changes, ensure compatibility with dependent services and tests.

### DIFF
--- a/signer/metrics.go
+++ b/signer/metrics.go
@@ -171,28 +171,32 @@ var (
 			"(Should not increase beyond block time; If high, may indicate raft joining issue for CoSigner) ",
 	})
 
-	missedPrecommits = promauto.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "signer_missed_precommits",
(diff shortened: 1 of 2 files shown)


agouin:
Thank you for this! Merged in #219 and will be included in the v4 release

PR #215, opened 2023-11-08
remote signer grpc server

agouin:
Add additional service to accept privval connections over grpc rather than connect out.
Existing functionality remains the same, this is not a breaking change.

GRPC service is enabled by setting a non-empty `grpcAddr` string in the config.yaml. This will be used as the listen address. Example: `:5555`, `0.0.0.0:5555`

Service API is minimal and excludes any comet types
```
service RemoteSigner {
	rpc PubKey (PubKeyRequest) returns (PubKeyResponse) {}
	rpc Sign(strangelove.horcrux.SignBlockRequest) returns (strangelove.horcrux.SignBlockResponse) {}
}
```

Other:
- Bump go to 1.21
- Switch to buf for proto-gen


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/strangelove-ventures/horcrux
 
-go 1.19
+go 1.21
```
bd21, line 3: bump the version in test/go.mod too?

PR #223, opened 2023-11-17
fix error checking over rpc

agouin:
The error detection with `errors.As` does not work for the RemoteCosigner since the error type is lost over grpc.

The purpose of this error check is to determine if there is a network error indicating the remote cosigner is down and the nonce cache should be invalidated for them. This can be replaced with a grpc status.Code check, where we only invalidate the cache on grpc errors that indicate an outage of the remote cosigner.

Improves a few logs


misko9 (approved):
much cleaner

PR #224, opened 2023-11-18
Fix - running out of nonces under load

agouin:
Account for continued nonce consumption in target

Sort by expiration when getting next nonce for signing so that soonest expiring nonces are consumed first.

Also fixed a bug with the same UUID being used for multiple sign operations (and failing for all after first) due to not locking for entire get>delete operation

### DIFF
--- a/signer/cosigner_nonce_cache.go
+++ b/signer/cosigner_nonce_cache.go
@@ -301,8 +301,8 @@ func (cnc *CosignerNonceCache) Start(ctx context.Context) {
 }
 
 func (cnc *CosignerNonceCache) GetNonces(fastestPeers []Cosigner) (*CosignerUUIDNonces, error) {
-	cnc.cache.mu.RLock()
-	defer cnc.cache.mu.RUnlock()
+	cnc.cache.mu.Lock()
+	defer cnc.cache.mu.Unlock()
 CheckNoncesLoop:
 	for _, cn := range cnc.cache.GetSortedByExpiration() {
 		var nonces CosignerNonces
@@ -321,9 +321,7 @@ CheckNoncesLoop:
 			}
 		}
 
-		cnc.cache.mu.RUnlock()
-		cnc.clearNonce(cn.UUID)
-		cnc.cache.mu.RLock()
+		delete(cnc.cache.cache, cn.UUID)
 
 		// all peers found
 		return &CosignerUUIDNonces{
@@ -348,15 +346,10 @@ func (cnc *CosignerNonceCache) PruneNonces() {
 	defer cnc.cache.mu.Unlock()
 	for u, cn := range cnc.cache.cache {
 		if time.Now().After(cn.Expiration) {
+			fmt.Printf("Pruning nonce %s\n", u)
 			delete(cnc.cache.cache, u)
 		}
 	}
-}
-
-func (cnc *CosignerNonceCache) clearNonce(uuid uuid.UUID) {
-	cnc.cache.mu.Lock()
-	defer cnc.cache.mu.Unlock()
-	delete(cnc.cache.cache, uuid)
 }
 
 func (cnc *CosignerNonceCache) ClearNonces(cosigner Cosigner) {
--- a/signer/cosigner_nonce_cache.go
+++ b/signer/cosigner_nonce_cache.go
@@ -163,12 +163,6 @@ func (cnc *CosignerNonceCache) reconcile(ctx context.Context) {
 	if noncesPerMin < 0 {
 		noncesPerMin = 0
 	}
-	cnc.logger.Debug("Reconciling nonces",
-		"remaining", remainingNonces,
-		"lastReconcileNonces", lastReconcileNonces,
-		"noncesPerMin", noncesPerMin,
-		"timeSinceLastReconcile", timeSinceLastReconcile,
-	)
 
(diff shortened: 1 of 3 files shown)


signer/cosigner_nonce_cache_test.go
```diff
@@ -45,6 +68,9 @@ func TestNonceCacheDemand(t *testing.T) {
 
 	cancel()
 
-	target := int(nonceCache.noncesPerMinute*.01) + 10
+	target := int(nonceCache.noncesPerMinute*.01) + int(nonceCache.noncesPerMinute/30) + 100
```
misko9, line 71: Can you put a comment for this logic? E.g. how does `int(nonceCache.noncesPerMinute*.01)` and `100` map to the `target` in the `reconcile` function?. Is this check for ensuring the cache doesn't grow unbounded?

signer/cosigner_nonce_cache.go
```diff
@@ -116,16 +134,23 @@ func NewCosignerNonceCache(
...
 		logger:            logger,
 		cache:             NewNonceCache(),
 		cosigners:         cosigners,
 		leader:            leader,
 		getNoncesInterval: getNoncesInterval,
 		getNoncesTimeout:  getNoncesTimeout,
 		threshold:         threshold,
+		pruner:            pruner,
 	}
+	if pruner == nil {
```
misko9, line 149: Add a comment that this is expected to be nil for now, but the unit test sets it with a mockPruner.

signer/cosigner_nonce_cache.go
```diff
@@ -88,6 +95,17 @@ func (nc *NonceCache) Set(uuid uuid.UUID, cn *CachedNonce) {
 	nc.cache[uuid] = cn
 }
 
+func (nc *NonceCache) GetSortedByExpiration() []*CachedNonce {
```
misko9, line 98: Should we add an enhancement issue to remove this sorting and change the CosignerNonceCache's NonceCache's cache from a map to an array which would be FIFO and pruning can be optimized?

signer/cosigner_nonce_cache.go
```diff
@@ -316,22 +341,17 @@ CheckNoncesLoop:
 	return nil, fmt.Errorf("no nonces found involving cosigners %+v", cosignerInts)
 }
 
-func (cnc *CosignerNonceCache) pruneNonces() {
+func (cnc *CosignerNonceCache) PruneNonces() {
 	cnc.cache.mu.Lock()
 	defer cnc.cache.mu.Unlock()
 	for u, cn := range cnc.cache.cache {
 		if time.Now().After(cn.Expiration) {
+			fmt.Printf("Pruning nonce %s\n", u)
```
misko9, line 349: Is this print intentional?

misko9 (approved):
Logic looks good, just some comments to address.

signer/cosigner_nonce_cache_test.go
agouin: It was supposed to be the same target as in the `reconcile` method. I had changed the `+ 10` to `+ 100` then back, but forgot to update the test. I pulled the `target` out to a method so it can be called in the test instead.

signer/cosigner_nonce_cache.go
agouin: yes, this makes sense. good call

signer/cosigner_nonce_cache.go
agouin: no, thanks. Removed.

signer/cosigner_nonce_cache.go
```diff
@@ -341,21 +323,27 @@ CheckNoncesLoop:
...
+func (cnc *CosignerNonceCache) PruneNonces() int {
 	cnc.cache.mu.Lock()
 	defer cnc.cache.mu.Unlock()
-	for u, cn := range cnc.cache.cache {
-		if time.Now().After(cn.Expiration) {
-			fmt.Printf("Pruning nonce %s\n", u)
-			delete(cnc.cache.cache, u)
+	nonExpiredIndex := len(cnc.cache.cache) - 1
+	for i := len(cnc.cache.cache) - 1; i >= 0; i-- {
+		if time.Now().Before(cnc.cache.cache[i].Expiration) {
```
misko9, line 331: If the whole cache expires, it doesn't look like anything would be pruned

signer/cosigner_nonce_cache.go
```diff
@@ -62,48 +60,28 @@ func (lc *lastCount) Get() int {
...
-	cns := make([]*CachedNonce, 0, len(nc.cache))
-	for _, cn := range nc.cache {
-		cns = append(cns, cn)
-	}
-	sort.Slice(cns, func(i, j int) bool {
-		return cns[i].Expiration.Before(cns[j].Expiration)
-	})
-	return cns
+func (nc *NonceCache) Delete(index int) {
+	nc.cache = append(nc.cache[:index], nc.cache[index+1:]...)
```
misko9, line 84: Does this succeed for both the first and last index in the slice?

agouin: yes, added unit test

signer/cosigner_nonce_cache.go
agouin: good catch, fixed and added test

misko9 (approved):

PR #229, opened 2023-12-01
Low nonce consumption

### LINKED ISSUE #228: Wrong "state" migrating from 3.1.0 to 3.2.0
Migrating from version 3.1.0 to 3.2.0 no signatures occur with this error

`Cosigner failed to set nonces and sign       chain_id=xxx-1 height=xxx round=0 type=prevote cosigner=1 err="unexpected state, metadata does not exist for U: xxxxx"`

The steps performed:

- stop all sentry/nodes
- stop all signers
- update bin
- rm -rf /raft
- start all signers
- start sentry/nodes

The same in test with a stand-alone setup (--accept-risk) with a wrong state file.
In this last case I solved it increasing blocks without signatures and then by deleting `state/priv_validator_state.json` (it's a testnet)

Some idea?
Thanks

agouin:
Fine tune when nonces are being consumed at a delay greater than the getNoncesInterval

When nonces are being consumed at a rate less than the getNoncesInterval, the resulting signal makes it difficult to estimate the correct target that will prevent expiring nonces and also prevent cached nonces dropping to zero.

This introduces a better moving average mechanism to more accurately determine the rate of nonce consumption, and enables reconciling to get more nonces as soon as the cache is empty.

Resolves #228

### DIFF
--- a/signer/cosigner_nonce_cache_test.go
+++ b/signer/cosigner_nonce_cache_test.go
@@ -287,5 +287,5 @@ func TestNonceCacheDemandSlowDefault(t *testing.T) {
 
 	cancel()
 
-	require.LessOrEqual(t, nonceCache.cache.Size(), nonceCache.target(nonceCache.movingAverage.average()))
+	require.LessOrEqual(t, nonceCache.cache.Size(), nonceCache.target(60/7))
 }


signer/cosigner_nonce_cache.go
```diff
@@ -36,6 +34,56 @@ type CosignerNonceCache struct {
...
+		m.items[:keep+1]...,
+	)
+}
+
+func (m *movingAverage) average() float64 {
+	weightedSum := float64(0)
+	duration := float64(0)
+
+	for _, e := range m.items {
+		d := float64(e.timeSinceLastReconcile)
```
misko9, line 81: Won't this result in returning nonces/nanosecond? Maybe this: `d := e.timeSinceLastReconcile.Minutes()`?

misko9: Actually, the scaling doesn't matter since it cancels itself out.

misko9 (approved):

PR #240, opened 2023-12-21
Update Horcrux go.mod module path

jakubbujny:
Horcrux is currently at major version 3.X
From golang perspective module path should be updated to contain `v3` in path, ref: https://go.dev/doc/modules/version-numbers#major 

> A major version update to a number higher than v1 will also have a new module path. That’s because the module path will have the major version number appended, as in the following example:
>module example.com/mymodule/v2 v2.0.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@ COMMIT  := $(shell git log -1 --format='%H')
 
 all: install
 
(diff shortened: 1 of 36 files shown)


agouin (approved):
Thank you!

PR #241, opened 2023-12-22
vendor necessary comet

agouin:
rather than fork comet, we can tree shake and bring in only the components needed by horcrux in order to support bn254.


nitronit:
Hey @agouin, I know you guys are putting a lot of effort in Horcrux but this looks pretty un-gophy and very anti-pattern TBH! 
And looks like a nightmare to support/debug etc etc. 

My take on this is: 
- The cosigners/mpc should never know or bother what it is trying to "organise".
- This is/should be the role of the threshold_validator (the server for the sentry/cometbft client). Basically a proxy in front of the MPC.  
- The same should almost apply for the MPC (cosigners) with regards to the tss.


Happy holidays!

agouin:
@nitronit I appreciate you taking the time to look through the horcrux PRs, I can assure you that this is very much an experiment of removing the comet dependency to avoid a fork and not representative of how things will be organized

PR #297, opened 2025-03-07
single lock for read and write

agouin:
patch rare race condition that allows a sign state read to occur multiple times prior to write, bypassing synchronous checks.


signer/sign_state_test.go
```diff
@@ -0,0 +1,104 @@
...
+	for i := 1; i < 100; i++ {
+		const jMax = 100
+		startedCh := make(chan struct{}, jMax)
+		workNowCh := make(chan struct{})
+
+		numErr, numSuccess := 0, 0
+		var mu sync.Mutex
+		i := i
+		var wg sync.WaitGroup
+		wg.Add(100)
```
mark-rushakoff, line 25: Could be more consistent about referencing the `jMax` constant here and in the counts in the final assertion. But OTOH this test may never end up changing anyway.

mark-rushakoff (approved):
This should be good to go. Getting the tests in place was key here.

PR #307, opened 2025-07-31
Feature/cometbft v1.0 support

qj0r9j0vc2:
### Description

  This PR adds support for CometBFT v1.0 protocol while maintaining backward compatibility with legacy (v0.38) protocol. This resolves the critical issue where validators using Horcrux encounter "unsupported key" errors when chains upgrade to CometBFT v1.0.

  ### Problem

  During the Injective v1.16.0 upgrade (CometBFT v0.38.17 → v1.0.1), validators using remote signing solutions encountered:
  error with private validator socket client: can't get pubkey: encoding: unsupported key

  This was due to breaking changes in the CometBFT privval protocol that weren't supported by existing remote signers.

  ### Solution

  Implemented a dual-protocol architecture that:
  - Supports both legacy (v0.38) and v1.0 protocol messages
  - Converts between protocol versions as needed
  - Maintains full backward compatibility with existing deployments
  - Adds configuration option to explicitly set protocol version

  ### Key Changes

  1. **Protocol Abstraction Layer** (`signer/privval_protocol.go`)
     - `PrivValProtocol` interface for protocol-agnostic handling
     - `LegacyProtocol` and `V1Protocol` implementations
     - Protocol selection based on configuration

  2. **Updated Remote Signer** (`signer/remote_signer_v2.go`)
     - New `ReconnRemoteSignerV2` supporting protocol version selection
     - Seamless handling of both protocol versions

  3. **Configuration** (`cmd/horcrux/cmd/config.go`)
     - Added `protocolVersion` field (defaults to "legacy" for compatibility)
     - Options: "legacy" (default) or "v1"

  4. **Protocol Differences Handled**:
     - Namespace: `tendermint.privval` → `cometbft.privval.v1`
     - PubKey encoding: wrapped `PublicKey` message → raw `pub_key_bytes` + `pub_key_type`
     - New fields: `skip_extension_signing` in SignVoteRequest

  5. **Bug Fix**: Fixed `proto.Equal()` comparison issue in `signer/file.go`
     - Replaced `proto.Equal()` with manual field comparison to avoid "proto: don't know how to compare 0" errors
     - Ensures timestamp reuse logic works correctly with both protocol versions

  ### Testing

  Comprehensive test suite added (`signer/privval_protocol_test.go`):
  - End-to-end protocol tests for both versions
  - Signature consistency verification
  - Edge case handling (nil values, extension signing)

Successfully tested on Injective testnet with CometBFT v1.0.

  ### Usage

  For chains using CometBFT v1.0 (like Injective v1.16.0+):
  ```yaml
 protocolVersion: v1

# For legacy chains (default, no config needed):
protocolVersion: legacy  # or omit entirely
```

###  Compatibility

  - Fully backward compatible - no breaking changes
  - Tested with both Injective (v1.0) and Axelar (legacy) scenarios
  - No action required for existing deployments (defaults to legacy)

###  Documentation

  Added comprehensive documentation in docs/cometbft-v1-migration.md explaining:
  - Configuration options
  - Migration steps
  - Troubleshooting guide
  - Compatibility matrix

### CODE REVIEW

File: docs/cometbft-v1-config-example.yaml [hunk 1]
```diff
@@ -0,0 +1,35 @@
+# Example Horcrux configuration for CometBFT v1.0 protocol support
+# This configuration demonstrates how to enable CometBFT v1.0 protocol compatibility
+
+# Signing mode - can be "single" or "threshold"
+signMode: single
+
+# Protocol version - REQUIRED for CometBFT v1.0 support
+# Options: "legacy" (default, for v0.38 and earlier) or "v1" (for v1.0+)
+protocolVersion: v1
```
agouin: This needs to be a config on the `chainNode`. One horcrux cluster can sign for multiple chains at once, and this will require separate clusters for chains < v1 and >= v1.

We should be able to support signing both with the same cluster

qj0r9j0vc2: I'll take it

qj0r9j0vc2: now protocol Version could be set by chainNode.

- [feat: per-node CometBFT protocol version support](https://github.com/qj0r9j0vc2/horcrux/commit/d000ce5805a22ac10e89694ae5f86d95d778cfbd)

qj0r9j0vc2 (commented):
this commit consists [Fix proto.Equal comparison issue causing false "conflicting data" errors #2](https://github.com/qj0r9j0vc2/horcrux/pull/2)

go.mod
```diff
@@ -1,18 +1,21 @@
 module github.com/strangelove-ventures/horcrux/v3
 
-go 1.21
+go 1.23.1
+
+toolchain go1.24.2
```
aljo242: let's remove this

signer/threshold_validator.go
```diff
@@ -451,8 +473,8 @@ func (pv *ThresholdValidator) compareBlockSignatureAgainstSSC(
 		return nil, nil, stamp, err
 	}
 
-	// only differ by timestamp, okay to sign again
-	return nil, nil, stamp, nil
+	// Ensures deterministic signatures for same H/R/S.
+	return existingSignature.Signature, existingSignature.VoteExtensionSignature, stamp, nil
```
agouin: I am weary of this change along with the rest of the changes in ThresholdValidator. cometbft <v1 expects deterministic signatures for same HRST. If we return a signature for a different timestamp, it will not be valid, and therefore we have hindered the high availability qualities of horcrux. Why is this necessary for cometbft v1?

qj0r9j0vc2: I agree with your opinion. I'll remove it for this PR.

cc. @nmarcetic

Issue #308, opened 2025-09-12
Potential Safety Vulnerability: Amnesia Fault Leading to Disagreement in Tendermint Consensus


ranchalp:
This issue outlines a potential safety vulnerability identified by the Sei Labs team that can affect Tendermint-based chains. The vulnerability arises from an "amnesia" fault in high-availability (HA) validator setups, such as those using Horcrux, which can theoretically lead to a chain fork (disagreement).

Tendermint's safety proofs rely on the assumption that a correct validator process does not "forget" its state within a consensus round, particularly its lock on a proposed block.

A Horcrux setup distributes the role of a single validator across multiple physical nodes. Under specific network partition conditions, these nodes can receive different subsets of consensus messages. This can lead to a fragmented collective state where the validator, as a whole, fails to maintain a lock it should have acquired. We refer to this as an amnesia fault. This fault allows the validator to vote for a conflicting block in a subsequent round, breaking Tendermint's safety guarantees.

**Detailed Scenario Leading to Disagreement**

Setup: The network is partitioned into three groups of validators: A, B, and C. Validator set A uses a Horcrux setup, controlling > 1/3 of the voting power, with power(B) = power(C).

- A network partition occurs. A proposed block, Block V, is voted on. Partition C receives a sufficient number of prevotes and precommits (2f+1) from A and C and successfully commits Block V.

- Partitions A and B do not see enough precommits to commit Block V. They time out and move to the next round. However, based on the prevotes it saw and sent, validator set A should have entered a locked state on Block V.

- Because validator set A consists of multiple Horcrux nodes that saw different subsets of messages, their collective state is inconsistent. The validator as a whole effectively "forgets" its lock on Block V.

- In the subsequent round, a new, conflicting block, Block W, is proposed. Since validator set A has amnesia and is no longer locked, it is free to vote for Block W, which commits with the help of validator set B.

Result: The network is now forked. Partition C has committed Block V, while Partitions A and B have committed Block W.

**Proposed Long-Term Solution & Discussion**

While we are implementing a protocol-level mitigation on Sei (an additional $f+1$ precommit rule), we believe the most robust, long-term solution lies in enhancing the validator tooling itself. The root cause is that the remote signer is stateless with respect to consensus locks.

[...]

PR #310, opened 2025-10-08
Fix amnesia faults vulnerability #308

ranchalp:
## Problem Description

This implementation addresses a critical safety vulnerability (#308 ) in Horcrux that can lead to chain forks in Tendermint-based networks. The vulnerability, known as the "amnesia fault," occurs when Horcrux nodes in a distributed validator setup lose track of their consensus locks due to network partitions, allowing them to sign conflicting blocks.

### The Amnesia Fault Scenario

1. **Network Partition**: The network splits into groups A, B, and C
2. **Partial Lock**: Group A sees enough prevotes to lock on Block V at Round R
3. **Missing Lock**: Group B doesn't see the prevotes and remains unlocked
4. **Network Recovery**: When the network recovers, Group B can sign a conflicting Block W
5. **Chain Fork**: This leads to a chain fork as different parts of the network have committed different blocks

### Root Cause

The root cause is that **Horcrux is stateless with respect to Tendermint consensus locks**. The remote signer only tracks what it has signed (high watermark) but doesn't maintain the consensus-critical state of what the validator is locked on. This allows the validator to "forget" its lock and sign conflicting blocks.

## Solution Overview

This implementation adds **consensus lock tracking** to Horcrux, making the remote signer consensus-aware. The solution ensures that:

1. **Lock Tracking**: The system tracks what block/value the validator is locked on
2. **Lock Validation**: Before signing any block, the system checks if it would violate an existing lock
3. **Lock Updates**: Locks are updated when PRECOMMIT messages are signed according to Tendermint rules
4. **Lock Persistence**: Locks persist for all future rounds within the same height

## Tendermint Locking Rules Implementation

The implementation follows the correct Tendermint consensus locking rules:

**If a PRECOMMIT for value V is signed in round R for tendermint's consensus instance Id then:**

1. **Rule 1.1 [lock on V]**: For PROPOSAL and PREVOTE messages, allow only signing for V in rounds R' ≥ R
2. **Rule 1.2 [lock on V']**: If a PRECOMMIT message for V' is requested and signed in round R' > R such that V' ≠ V, then lock on V' instead for all rounds R'' > R'

**Key Points:**
- Locks persist for ALL future rounds within the same height
- Locks are only cleared when moving to a different height
- PRECOMMIT messages in later rounds can release and set new locks
- PROPOSAL and PREVOTE messages in later rounds must respect the existing lock

#### ConsensusLock
```go
type ConsensusLock struct {
    Height    int64  `json:"height"`
    Round     int64  `json:"round"`
    Value     []byte `json:"value,omitempty"`     // The locked block hash/value
    ValueType string `json:"value_type"`         // "block" or "nil"
}
```

#### Updated SignState
The `SignState` structure now includes a `ConsensusLock` field to track the current consensus lock state.

#### ValidateConsensusLock
```go
func (signState *SignState) ValidateConsensusLock(hrs HRSKey, signBytes []byte) error
```
- Checks if signing the given block would violate an existing consensus lock
- Returns an error if the lock would be violated
- Allows signing if no lock exists or if the lock is not applicable

#### Lock Update Logic
The consensus lock is updated in three scenarios:
1. **First lock for height**: When no lock exists, set the lock based on PRECOMMIT
2. **Lock release/update**: When PRECOMMIT for different value V' is signed in higher round R' > R
3. **Lock preservation**: When PRECOMMIT for same value V is signed in higher round (no change needed)

#### ThresholdValidator.Sign()
- Added consensus lock validation before proceeding with signing
- Updates consensus lock when PRECOMMIT is signed
- Preserves existing lock state in sign state consensus

#### LocalCosigner.sign()
- Added consensus lock validation before local signing
- Updates consensus lock when PRECOMMIT is signed locally

#### SignState.blockDoubleSign()
- Added consensus lock validation as the first check
- Ensures lock violations are caught early in the signing process

### DIFF
--- a/signer/local_cosigner.go
+++ b/signer/local_cosigner.go
@@ -220,13 +220,15 @@ func (cosigner *LocalCosigner) sign(req CosignerSignRequest) (CosignerSignRespon
 	}
 
 	// Check for consensus lock violations before proceeding
-	if err := ccs.lastSignState.ValidateConsensusLock(hrst.HRSKey(), req.SignBytes); err != nil {
+	// Use POL round validation
+	if err := ccs.lastSignState.ValidateConsensusLock(hrst.HRSKey(), req.SignBytes, req.PolRound); err != nil {
 		// Log the specific consensus lock violation with context
 		cosigner.logger.Error("Consensus lock violation in local cosigner",
 			"chain_id", chainID,
 			"height", hrst.Height,
 			"round", hrst.Round,
 			"step", hrst.Step,
+			"pol_round", req.PolRound,
 			"error", err,
 		)
 		// Return the specific consensus lock violation error with full details
--- a/signer/local_cosigner.go
+++ b/signer/local_cosigner.go
@@ -1,7 +1,6 @@
 package signer
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -222,7 +221,16 @@ func (cosigner *LocalCosigner) sign(req CosignerSignRequest) (CosignerSignRespon
 
 	// Check for consensus lock violations before proceeding
 	if err := ccs.lastSignState.ValidateConsensusLock(hrst.HRSKey(), req.SignBytes); err != nil {
-		return res, fmt.Errorf("consensus lock violation: %w", err)
+		// Log the specific consensus lock violation with context
+		cosigner.logger.Error("Consensus lock violation in local cosigner",
+			"chain_id", chainID,
+			"height", hrst.Height,
+			"round", hrst.Round,
+			"step", hrst.Step,
+			"error", err,
+		)
+		// Return the specific consensus lock violation error with full details
+		return res, err
 	}
 
 	// This function has multiple exit points.  Only start time can be guaranteed
@@ -298,45 +306,7 @@ func (cosigner *LocalCosigner) sign(req CosignerSignRequest) (CosignerSignRespon
 	}
 
 	// Handle consensus lock updates according to Tendermint rules
-	if hrst.Step == stepPrecommit {
-		// Extract the block hash from the sign bytes
-		blockHash := extractBlockHashFromSignBytes(req.SignBytes, hrst.Step)
-		if blockHash != nil {
-			// Rule 1.2: If PRECOMMIT for V' is signed in round R' > R where V' != V,
-			// then lock on V' instead for all rounds R'' > R'
-			if hrst.Round > ccs.lastSignState.ConsensusLock.Round &&
-				ccs.lastSignState.ConsensusLock.IsLocked() &&
-				!bytes.Equal(blockHash, ccs.lastSignState.ConsensusLock.Value) {
-				// Release old lock and set new lock on V'
-				signStateConsensus.ConsensusLock = ConsensusLock{
-					Height:    hrst.Height,
-					Round:     hrst.Round,
-					Value:     blockHash,
-					ValueType: "block",
-				}
-			} else if !ccs.lastSignState.ConsensusLock.IsLocked() {
-				// First lock for this height
-				signStateConsensus.ConsensusLock = ConsensusLock{
-					Height:    hrst.Height,
-					Round:     hrst.Round,
-					Value:     blockHash,
-					ValueType: "block",
-				}
-			} else {
-				// If PRECOMMIT for same value V in higher round, keep existing lock (no change needed)
-				signStateConsensus.ConsensusLock = ccs.lastSignState.ConsensusLock
-			}
-		}
-	} else {
-		// For non-PRECOMMIT steps, only clear lock if moving to different height
-		// Locks persist for all future rounds within the same height
-		if hrst.Height != ccs.lastSignState.ConsensusLock.Height {
-			signStateConsensus.ConsensusLock = ConsensusLock{}
-		} else {
-			// Preserve existing lock
-			signStateConsensus.ConsensusLock = ccs.lastSignState.ConsensusLock
-		}
-	}
+	signStateConsensus.ConsensusLock = updateConsensusLock(ccs.lastSignState.ConsensusLock, hrst.HRSKey(), req.SignBytes)
 
 	err = ccs.lastSignState.Save(signStateConsensus, &cosigner.pendingDiskWG)
 
--- a/signer/local_cosigner.go
+++ b/signer/local_cosigner.go
@@ -306,7 +306,7 @@ func (cosigner *LocalCosigner) sign(req CosignerSignRequest) (CosignerSignRespon
 	}
 
 	// Handle consensus lock updates according to Tendermint rules
-	signStateConsensus.ConsensusLock = updateConsensusLock(ccs.lastSignState.ConsensusLock, hrst.HRSKey(), req.SignBytes)
+	signStateConsensus.ConsensusLock = nextConsensusLock(ccs.lastSignState.ConsensusLock, hrst.HRSKey(), req.SignBytes)
 
 	err = ccs.lastSignState.Save(signStateConsensus, &cosigner.pendingDiskWG)
 
--- a/signer/local_cosigner.go
+++ b/signer/local_cosigner.go
@@ -1,6 +1,7 @@
 package signer
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -219,6 +220,11 @@ func (cosigner *LocalCosigner) sign(req CosignerSignRequest) (CosignerSignRespon
 		return res, err
 	}
 
+	// Check for consensus lock violations before proceeding
+	if err := ccs.lastSignState.ValidateConsensusLock(hrst.HRSKey(), req.SignBytes); err != nil {
+		return res, fmt.Errorf("consensus lock violation: %w", err)
+	}
+
 	// This function has multiple exit points.  Only start time can be guaranteed
 	metricsTimeKeeper.SetPreviousLocalSignStart(time.Now())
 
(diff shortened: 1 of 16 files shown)


signer/sign_state.go
```diff
@@ -108,6 +121,9 @@ type SignState struct {
 	SignBytes              cometbytes.HexBytes `json:"signbytes,omitempty"`
 	VoteExtensionSignature []byte              `json:"vote_ext_signature,omitempty"`
 
+	// Consensus lock tracking to prevent amnesia faults
+	ConsensusLock ConsensusLock `json:"consensus_lock,omitempty"`
```
pompon0: omitzero? omitempty does not apply to custom types: https://pkg.go.dev/encoding/json

signer/sign_state.go
```diff
@@ -98,6 +98,19 @@ func StepToType(step int8) cometproto.SignedMsgType {
...
+type ConsensusLock struct {
+	Height    int64  `json:"height"`
+	Round     int64  `json:"round"`
+	Value     []byte `json:"value,omitempty"` // The locked block hash/value
+	ValueType string `json:"value_type"`      // "block" or "nil"
+}
+
+// IsLocked returns true if there is an active consensus lock
+func (lock *ConsensusLock) IsLocked() bool {
+	return lock.Height > 0 && lock.Round >= 0
```
pompon0: it should be possible to lock at height 0 and round 0, no?

signer/sign_state.go
```diff
@@ -98,6 +98,19 @@ func StepToType(step int8) cometproto.SignedMsgType {
 	}
 }
 
+// ConsensusLock represents a Tendermint consensus lock on a specific value
+type ConsensusLock struct {
+	Height    int64  `json:"height"`
+	Round     int64  `json:"round"`
+	Value     []byte `json:"value,omitempty"` // The locked block hash/value
+	ValueType string `json:"value_type"`      // "block" or "nil"
```
pompon0: how about having just Value? nil for nil locked, non-nil for block locked.

signer/sign_state.go
```diff
@@ -577,6 +613,152 @@ func newDiffBlockIDsError(first, second []byte) *DiffBlockIDsError {
 	}
 }
 
+// ConsensusLockViolationError represents an error when trying to sign a block that violates a consensus lock
+type ConsensusLockViolationError struct {
+	msg string
```
pompon0: nit: you can quickly define a custom error types as `type ErrName { error }` and then construct it using `ErrName{fmt.Errorf(...)}`. And the most proper way of defining structured errors is `type ErrName { <custom fields> }` and postponing the fmt.Sprintf to `func (e *ErrName) Error() { return fmt.Sprintf() }`, so that message rendering can be skipped altogether if not needed.

signer/sign_state.go
```diff
@@ -577,6 +613,152 @@ func newDiffBlockIDsError(first, second []byte) *DiffBlockIDsError {
...
+func IsConsensusLockViolationError(err error) bool {
+	_, ok := err.(*ConsensusLockViolationError)
+	return ok
+}
+
+// IsConsensusLockStepViolationError checks if the error is a consensus lock step violation
+func IsConsensusLockStepViolationError(err error) bool {
+	_, ok := err.(*ConsensusLockStepViolationError)
+	return ok
+}
```
pompon0, line 710: nit: the canonical way to casting errors is by using `errors.As()`.

signer/sign_state.go
```diff
@@ -577,6 +613,152 @@ func newDiffBlockIDsError(first, second []byte) *DiffBlockIDsError {
...
+
+// extractBlockHashFromSignBytes extracts the block hash from Tendermint sign bytes
+func extractBlockHashFromSignBytes(signBytes []byte, _ int8) []byte {
+	// This is a simplified implementation - in practice, you'd need to properly
+	// unmarshal the Tendermint vote/proposal structure to extract the block hash
+	// For now, we'll use a placeholder that returns the first 32 bytes as a hash
+	if len(signBytes) < 32 {
+		return nil
+	}
```
pompon0, line 816: that needs to be fixed before merging, no?

pompon0: and we need an e2e test that it actually works

ranchalp: Done thanks.

signer/sign_state.go
ranchalp: fair, done. Thanks

signer/sign_state.go
ranchalp: It will show up as an empty struct {} when not locked now.

signer/sign_state.go
ranchalp: Fair, done thanks

signer/sign_state.go
ranchalp: yes, thanks fixed.

signer/sign_state.go
ranchalp: Done thanks.

signer/sign_state.go
ranchalp: Done thanks

ranchalp:
Thanks @pompon0 I addressed your comments

signer/local_cosigner.go
```diff
@@ -280,14 +294,21 @@ func (cosigner *LocalCosigner) sign(req CosignerSignRequest) (CosignerSignRespon
...
 		Step:                   hrst.Step,
 		Signature:              sig,
 		SignBytes:              req.SignBytes,
 		VoteExtensionSignature: res.VoteExtensionSignature,
-	}, &cosigner.pendingDiskWG)
+		ConsensusLock:          ccs.lastSignState.ConsensusLock, // Preserve existing lock
+	}
+
+	// Handle consensus lock updates according to Tendermint rules
+	signStateConsensus.ConsensusLock = updateConsensusLock(ccs.lastSignState.ConsensusLock, hrst.HRSKey(), req.SignBytes)
```
pompon0: it is misleading that you are setting ConsensusLock to the old value and then immediately overwriting it. inline the updateConsensusLock call instead.

pompon0: the "update" prefix is implying that updateConsensusLock is mutating its argument(s). Consider something like "nextConsensusLock" or some other name which would imply that it is a pure function.

signer/sign_state.go
```diff
@@ -577,6 +611,183 @@ func newDiffBlockIDsError(first, second []byte) *DiffBlockIDsError {
...
+	defer signState.mu.RUnlock()
+
+	// If no consensus lock exists, allow signing
+	if !signState.ConsensusLock.IsLocked() {
+		return nil
+	}
+
+	// If we're signing for a different height, the lock is no longer relevant
+	if hrs.Height != signState.ConsensusLock.Height {
+		return nil
```
pompon0, line 725: nit: `hrs.Height > ...` for defense-in-depth

pompon0: and return error if `<`

pompon0: (same for round)

signer/sign_state.go
```diff
@@ -577,6 +611,183 @@ func newDiffBlockIDsError(first, second []byte) *DiffBlockIDsError {
...
+	}
+
+	// For PROPOSAL and PREVOTE messages in rounds R' >= R, only allow signing for the locked value V
+	if (hrs.Step == stepPropose || hrs.Step == stepPrevote) && hrs.Round >= signState.ConsensusLock.Round {
+		// Extract the block hash from the sign bytes to compare with the locked value
+		blockHash := extractBlockHashFromSignBytes(signBytes, hrs.Step)
```
pompon0: make it return an error if parsing fails

signer/sign_state.go
```diff
@@ -577,6 +611,183 @@ func newDiffBlockIDsError(first, second []byte) *DiffBlockIDsError {
...
+		if blockHash == nil {
+			// If we can't extract the block hash, allow signing (fallback to existing behavior)
+			return nil
```
pompon0: imo, it is more desirable to fail when we don't know what we are signing.

signer/sign_state.go
```diff
@@ -577,6 +611,183 @@ func newDiffBlockIDsError(first, second []byte) *DiffBlockIDsError {
...
+		return nil
+	}
+
+	// If we're signing for a different height, the lock is no longer relevant
+	if hrs.Height != signState.ConsensusLock.Height {
+		return nil
+	}
+
+	// For PROPOSAL and PREVOTE messages in rounds R' >= R, only allow signing for the locked value V
+	if (hrs.Step == stepPropose || hrs.Step == stepPrevote) && hrs.Round >= signState.ConsensusLock.Round {
```
pompon0, line 729: I think we have a problem here - we should compare pol_round against ConsensusLock.Round here, not current round. For proposal signing we can extract it, but prevotes do not include pol round.

ranchalp:
Thanks for the comments, just addressed them, except for pushback on one of them (see the conversation in the related comment).

signer/sign_state.go
ranchalp: This is Tendermint type of stuff. What matters for Horcrux signing is consistency, which is only relative to the round of the latest signed precommit. That is, one thing is the highest round where Tendermint state sees the highest POL. But for singing on Horcrux what matters is that whatever is proposed for signing next is consistent with what has been proposed for signing before, not the POL. And the only time that we need to update and carry over locks when it comes to signing is after a Precommit message. 

This implementation of Horcrux should not conflict with a correct Tendermint signer since nodes will lock on a value only when sending a Precommit message, and only will sign Prevotes or send Proposals for the locked value until the next Precommit message (if the Tendermint implementation follows the Tendermint spec from [this paper](https://arxiv.org/pdf/1807.04938) as the official documentation says).

pompon0: discussed offline. We need to amend API to handle branch on line 28 of https://arxiv.org/pdf/1807.04938

ranchalp: https://github.com/strangelove-ventures/horcrux/pull/310#issuecomment-3436220042

signer/sign_state.go
pompon0: exactly, with omitzero it won't show up at all.

signer/sign_state.go
```diff
@@ -121,7 +150,11 @@ type SignState struct {
 	VoteExtensionSignature []byte              `json:"vote_ext_signature,omitempty"`
 
 	// Consensus lock tracking to prevent amnesia faults
-	ConsensusLock ConsensusLock `json:"consensus_lock"`
+	ConsensusLock ConsensusLock `json:"consensus_lock,omitempty"`
+
+	// Prevote quorums tracking for Tendermint Algorithm Line 28
+	// Map: height -> (value -> oldest round that received 2f+1 prevotes)
+	PrevoteQuorums map[string]map[string]int64 `json:"prevote_quorums"` // height -> (value -> round)
```
pompon0: why do you use string to index by height?

ranchalp:
Reverted to the version we discussed in sync that uses the POL round.

ericjohncarlson:
Been watching this PR closely and it seems like this has stalled out.  Is the original concern still a problem and is this expected to be merged?

PR #316, opened 2026-05-18
Fix Import math package for leader range checks

bjormgyg:
Add math import for leader validation in GetLeader method, Use an explicit `int32` range check before converting `leader` in `CosignerGRPCServer.GetLeader`.   Best minimal fix (without changing existing behavior for valid values): if `leader` is outside `[math.MinInt32, math.MaxInt32]`, return a safe fallback (`-1`, consistent with existing “unknown leader” sentinel), otherwise cast to `int32`.

Changes needed:
- `signer/cosigner_grpc_server`
  - Add `math` import.
  - Update `GetLeader` method to perform bounds check before `int32(leader)` conversion.

If a string is parsed into an int using `strconv.Atoi`, and subsequently that int is converted into another integer type of a smaller size, the result can produce unexpected values.

This also applies to the results of `strconv.ParseInt` and `strconv.ParseUint` when the specified size is larger than the size of the type that number is converted to.

**References**
[doc strconv.Atoi](https://golang.org/pkg/strconv/#Atoi)
[doc strconv.ParseInt](https://golang.org/pkg/strconv/#ParseInt)
[doc strconv.ParseUint](https://golang.org/pkg/strconv/#ParseUint)

### DIFF
--- a/signer/cosigner_grpc_server.go
+++ b/signer/cosigner_grpc_server.go
@@ -3,6 +3,7 @@ package signer
 import (
 	"context"
 	"fmt"
+	"math"
 
 	"github.com/google/uuid"
 	"github.com/hashicorp/raft"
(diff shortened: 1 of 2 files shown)


bjormgyg (commented):
Use an explicit `int`-range and validity check immediately after parsing, before any `int(pid)` conversion. Keep `ParseInt(..., 64)` if desired, but guard conversion with constants from `math`:
- require `pid > 0`
- require `pid <= math.MaxInt32` (portable safe upper bound for `int` on both 32-bit and 64-bit)

Then perform `pidInt := int(pid)` once and reuse it for both `os.Getpid()` comparison and `os.FindProcess`.

